PR #1, opened 2016-04-04
Fix compillation with 1.7.0 and nightly.

SiegeLord:
- Update version of bindgen, as 0.15 depended on a syntex version that just didn't compile for me.
- Because of the update to bindgen, fix various type issues.


adamcrume:
Thanks!

PR #2, opened 2016-04-05
Wrap TF_Run.

SiegeLord:
I've chosen an API that's quite a bit different from the C++ API because you've made your tensors typed.

I haven't added any tests as it wasn't immediately obvious how to do it, and I wanted your opinion on the API first. It seems to work on my graphs, however.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -12,6 +12,7 @@ use std::fmt;
 use std::fmt::Debug;
 use std::fmt::Display;
 use std::fmt::Formatter;
+use std::marker;
 use std::mem;
 use std::ops::Drop;
 use std::os::raw;
@@ -296,6 +297,46 @@ impl Session {
     }
     status
   }
+
+  pub fn run(&mut self, step: &mut Step) -> Status {
+    // Copy the input tensors because TF_Run consumes them.
+    let mut input_tensors = Vec::with_capacity(step.input_tensors.len());
+    for &input_tensor in &step.input_tensors {
+      let input_tensor = input_tensor as *const tf::TF_Tensor;
+      unsafe {
+        let mut dims = Vec::with_capacity(tf::TF_NumDims(input_tensor) as usize);
+        for i in 0..dims.capacity() {
+          dims.push(tf::TF_Dim(input_tensor, i as i32));
+        }
+        input_tensors.push(tf::TF_NewTensor(
+          tf::TF_TensorType(input_tensor),
+          dims.as_ptr() as *mut i64,
+          dims.len() as libc::c_int,
+          tf::TF_TensorData(input_tensor),
+          tf::TF_TensorByteSize(input_tensor),
+          Some(noop_deallocator),
+          std::ptr::null_mut()));
+      }
+    }
+
+    let status = Status::new();
+    unsafe {
+      tf::TF_Run(
+        self.inner,
+        std::ptr::null(),
+        step.input_name_ptrs.as_mut_ptr(),
+        input_tensors.as_mut_ptr(),
+        input_tensors.len() as raw::c_int,
+        step.output_name_ptrs.as_mut_ptr(),
+        step.output_tensors.as_mut_ptr(),
+        step.output_tensors.len() as raw::c_int,
+        step.target_name_ptrs.as_mut_ptr(),
+        step.target_name_ptrs.len() as raw::c_int,
+        std::ptr::null_mut(),
+        status.inner);
+    };
+    status
+  }
 }
 
 impl Drop for Session {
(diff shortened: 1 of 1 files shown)


SiegeLord:
Here's what the API looks like in use, if it's not obvious:

```rust
extern crate tensorflow;

use tensorflow as tf;
use std::fs::File;
use std::io::Read;

fn main()
{
	let opts = tf::SessionOptions::new();
	let mut sess = tf::Session::new(&opts).unwrap();
	
	let mut f = File::open("model.pb").unwrap();
	let mut buf = vec![];
	f.read_to_end(&mut buf).unwrap();
	let status = sess.extend_graph(&buf);
	println!("{}", status);
	
	let mut a = tf::Tensor::<f32>::new(&[1]);
	a.data_mut()[0] = 1.0;
	let mut b = tf::Tensor::<f32>::new(&[1]);
	b.data_mut()[0] = 2.0;
	
	let mut step = tf::Step::new();
	step.feed("a:0", &a).unwrap();
	step.feed("b:0", &b).unwrap();
	let output = step.fetch("out:0").unwrap();
	
	let status = sess.run(&mut step);
	println!("{}", status);
	
	println!("{}", step.take_output::<f32>(output).unwrap().data()[0]);
}
```

(side note, it'd be a lot more convenient if the `Status` returning functions actually returned Result<(), Status> instead so you could `try!` and `unwrap` them).

adamcrume:
I'd rather avoid exposing raw pointers and the raw FFI in the Rust API if at all possible.  Take a look at https://github.com/google/tensorflow-rust/tree/wip.  It currently fails with "Invalid argument: You must feed a value for placeholder tensor 'x' with dtype float [[Node: x = Placeholder[dtype=DT_FLOAT, shape=[], _device="/job:localhost/replica:0/task:0/cpu:0"]()]]" for some reason, even though I'm passing in a float tensor with name x.

SiegeLord:
Hmm, I don't think I expose any raw pointers, do I? My initial attempt also used a type just like your `AnyTensor` but I decided it was too annoying to convert back and forth from it.

SiegeLord:
One advantage of my API is that's it's easy to use the same tensor multiple times in a single run call. It's probably possible to do something like that with `AnyTensor`, but you'd have to create some aliasing support.

One disadvantage of my API is that it's not easy to reuse the allocations inside `Step` between multiple run calls. Not sure how important this is.

adamcrume:
You're right, of course; you don't seem to be exposing any raw pointers.  I must have misread your code the first time around.  I originally wanted to stick as closely as possible to the C API, but that's probably not the best way to handle things.  I like the way that you can use Step::feed instead of trying to build a heterogeneous slice (which is why I created AnyTensor).

If you make a few changes, I'd be happy to merge it:

- Check the tensor type in take_output and return error if it doesn't match what the user is requesting.
- Return a Result from take_output.  For consistency, you could create a Status with Status::new_set and return that on error.  I'm fine with using something else, though, especially since that would avoid round-tripping from a Rust string to a C string and back.
- Rename Step::fetch to something a little more helpful. Maybe reserve_output or request_output?
- Add the test below.

```
  #[test]
  fn test_run() {
    // Graph is just y = 2 * x                                                                                                                                                                
    let graph_proto = vec![
      0x0a, 0x2a, 0x0a, 0x01, 0x78, 0x12, 0x0b, 0x50, 0x6c, 0x61, 0x63, 0x65, 0x68, 0x6f, 0x6c, 0x64,
      0x65, 0x72, 0x2a, 0x0b, 0x0a, 0x05, 0x64, 0x74, 0x79, 0x70, 0x65, 0x12, 0x02, 0x30, 0x01, 0x2a,
      0x0b, 0x0a, 0x05, 0x73, 0x68, 0x61, 0x70, 0x65, 0x12, 0x02, 0x3a, 0x00, 0x0a, 0x30, 0x0a, 0x03,
      0x79, 0x2f, 0x79, 0x12, 0x05, 0x43, 0x6f, 0x6e, 0x73, 0x74, 0x2a, 0x0b, 0x0a, 0x05, 0x64, 0x74,
      0x79, 0x70, 0x65, 0x12, 0x02, 0x30, 0x01, 0x2a, 0x15, 0x0a, 0x05, 0x76, 0x61, 0x6c, 0x75, 0x65,
      0x12, 0x0c, 0x42, 0x0a, 0x08, 0x01, 0x12, 0x00, 0x2a, 0x04, 0x00, 0x00, 0x00, 0x40, 0x0a, 0x19,
      0x0a, 0x01, 0x79, 0x12, 0x03, 0x4d, 0x75, 0x6c, 0x1a, 0x01, 0x78, 0x1a, 0x03, 0x79, 0x2f, 0x79,
      0x2a, 0x07, 0x0a, 0x01, 0x54, 0x12, 0x02, 0x30, 0x01
    ];
    let mut session = create_session();
    let status = session.extend_graph(&graph_proto);
    assert!(status.is_ok());
    let mut x = <Tensor<f32>>::new(&[2]);
    x.data_mut()[0] = 2.0;
    x.data_mut()[1] = 3.0;
    let mut step = Step::new();
    step.feed("x:0", &x).unwrap();
    let output_ix = step.fetch("y:0").unwrap();
    let result = session.run(&mut step);
    assert!(result.is_ok(), "{}", result);
    let output_tensor = step.take_output::<f32>(output_ix).unwrap();
    let data = output_tensor.data();
    assert_eq!(data.len(), 2);
    assert_eq!(data[0], 4.0);
    assert_eq!(data[1], 6.0);
  }
```

I'll change function return values from Status to Result<(), Status> afterword.

SiegeLord:
I've made the changes you suggested. I went with using `Result` as is. It could always be optimized to not do the round-trip if that seems uncouth.

PR #15, opened 2016-07-23
Reorganize the bindings in the sys crate

IvanUkhov:
Anticipating the graph API that will be shipped with the next release of TensorFlow, I’ve started to reorganize the bindings in the sys crate in order to make the maintenance easier and the future extension more systematic. The types are now alphabetically sorted, and the functions are now grouped by the data structures that there are concerned with.

In addition, I’ve proactively made the following changes:

* use `int64_t` instead of `c_longlong`;
* change `TF_NewTensor`’s `dims` from `*mut _` to `*const _`; and
* change `TF_PRunSetup`’s `handle` from `*mut *mut _` to `*mut *const _`.

All of them are perfectly valid with the current version of TensorFlow, version 0.9. You can see individual commits in the [commit history](https://github.com/stainless-steel/tensorflux-sys/commits/master) of `tensorflux-sys`.

Regards,
Ivan

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -10,7 +10,7 @@ repository = "https://github.com/google/tensorflow-rust"
 
 [dependencies]
 libc = "0.2"
-tensorflow-sys = { version = "0.4", path = "tensorflow-sys" }
+tensorflow-sys = { version = "0.4.1", path = "tensorflow-sys" }
 
 [dev-dependencies]
 random = "0.12"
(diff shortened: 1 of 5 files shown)


IvanUkhov:
Based on Travis’ logs, it seems there is some cache-related problem when it gets to the execution of `cargo doc` in the root directory. It might be due to the fact that `tensorflow` refers to `tensorflow-sys` using a local path. In this scenario, Cargo gets confused where `target` is cached.

IvanUkhov:
The problem is actually due to a bug in Cargo’s workspaces. See issue [2897](https://github.com/rust-lang/cargo/issues/2897).

adamcrume:
I cleared the cache and restarted the build.  Hopefully that fixes the issue for now.

PR #20, opened 2016-09-17
Depend on num-complex instead of num

IvanUkhov:
By depending on `num-complex` directly, one can eliminate the need to download and compile such redundant crates as `num-bigint`, `num-integer`, `num-iter`, `num-rational`, and `rustc-serialize`.

Regards,
Ivan


adamcrume:
Thanks!

PR #43, opened 2016-12-21
Upgrade to TensorFlow 0.12.0

adamcrume:


adamcrume:
Fixes https://github.com/tensorflow/rust/issues/31, and then some.

travis-ci/install.sh
```diff
@@ -1,6 +1,6 @@
 #!/bin/bash
 
-BAZEL_URL=https://github.com/bazelbuild/bazel/releases/download/0.3.0/bazel-0.3.0-installer-linux-x86_64.sh
+BAZEL_URL=https://github.com/bazelbuild/bazel/releases/download/0.3.2/bazel-0.3.2-installer-linux-x86_64.sh
```
jhseu: Use the 0.4.3 release? I've personally run into strange bugs on the 0.3.x release.

src/graph.rs
```diff
@@ -126,6 +151,71 @@ impl Graph {
...
+
+  /// Returns the shape of the Tensor referenced by `output`.
+  ///
+  /// If the number of dimensions in the shape is unknown, the return value is
+  /// None. Otherwise, each element of `dims` will be set corresponding to the
+  /// size of the dimension. An unknown dimension is represented by `-1`.
+  ///
+  /// Returns an error if:
+  ///   * `output` is not in `graph`.
+  pub fn tensor_shape(&self, output: Output) -> Result<Option<Vec<i64>>> {
```
jhseu: I'd favor returning TensorShape type where fetching a dimension is Option<i64> and is None if the shape of that dim is unknown.

adamcrume: So something roughly like `type TensorShape = Vec<Option<i64>>` and have `tensor_shape()` return `Result<Option<TensorShape>>`?

jhseu:
`Result<TensorShape>` because a shape with an unknown number of dimensions is still valid.

TensorShape would have two functions (naming up to you):
`dim(i) -> Option<int64>`
`num_dims() -> Option<int64>`

travis-ci/install.sh
adamcrume: Easy enough. :)

adamcrume:
I added a `TensorShape` type which is the parameter type for `OperationDescription::set_attr_shape()` and the return type for `Graph::tensor_shape()`

src/lib.rs
```diff
@@ -1022,14 +1021,78 @@ trait OperationTrait {
...
+
+/// A TensorShape is the shape of a tensor.  A TensorShape may be an unknown
+/// rank, or it may have a known rank with each dimension being known or
+/// unknown.
+#[derive(Debug,Eq,Ord,PartialEq,PartialOrd,Hash,Clone)]
+pub struct TensorShape(Option<Vec<Option<i64>>>);
+
+impl TensorShape {
+  /// Returns the number of dimensions if known, or None if unknown.
+  pub fn rank(&self) -> Option<usize> {
```
jhseu: We're generally discouraging the use of the term `rank`:
https://www.tensorflow.org/how_tos/documentation/
`Refer to tensors with 3 or more dimensions as 3-D tensors or n-D tensors. Use the word "rank" only if it makes sense, but try to use "dimension" instead. Never use the word "order" to describe the size of a tensor.`

In the C++ class, we're using the convention dims() and dim_size().
https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/tensor_shape.h

adamcrume: Renamed to dims().  Out of curiosity, do you know why "rank" is discouraged?  Is it just to keep the jargon to a minimum?

PR #48, opened 2017-01-23
Correct bazel version parsing in build.rs

Enet4:
`tensorflow-sys` was failing to build in my system because apparently Bazel can report something like below, which was not being properly parsed by the building script.

```
Build label: 0.4.3- (@non-git)
Build target: bazel-out/local-opt/bin/src/main/java/com/google/devtools/build/lib/bazel/BazelServer_deploy.jar
Build time: Mon Jan 23 11:31:26 2017 (1485171086)
Build timestamp: 1485171086
Build timestamp as int: 1485171086
```

This PR will make version checking resistant to additional non-whitespace text after the version, as well as a sole trailing hyphen (which is not compliant to semver and would also output a parsing error). I also took the liberty of fixing a typo in the related error message.

### DIFF
--- a/tensorflow-sys/build.rs
+++ b/tensorflow-sys/build.rs
@@ -47,7 +47,7 @@ fn main() {
         log!("{:?} already exists, not building", library_path);
     } else {
         if let Err(e) = check_bazel() {
-            println!("cargo:error=Bazel must be installed at be version {} or greater. (Error: {})", MIN_BAZEL, e);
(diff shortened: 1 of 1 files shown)


Enet4:
> Once you've signed, please reply here (e.g. I signed it!) and we'll verify. Thanks.

I signed it, please verify now. :)

adamcrume:
Thanks, and sorry for the delay.  Got to figure out why the build is failing. :(

PR #53, opened 2017-02-24
Bump to tensorflow 1.0.0

daschl:
This changeset bumps the tensorflow dependency
to 1.0.0 and along the way gets rid of the zlib
patch which I think is not needed anymore.

@adamcrume let me know if there is stuff I missed, tests pass if I comment out the expression.rs example which did fail before on master. The patch prevent tf from compiling but removing it seems to work fine.


daschl:
Oh btw I have another PR pending which fixes the test failures by disabling the example when the experimental feature flag is nof enabled. Will put it up later

adamcrume:
Just for the record, this fixes #55.

adamcrume:
Thanks!

PR #56, opened 2017-02-26
Use libtensorflow.so instead of libtensorflow_c.so.

asimshankar:
The two are identical, but the former is what is recommended
(and used by TensorFlow team to build binary releases of the TensorFlow
C library)

The existence of both targets is an unfortunate outcome of the near-simultaneous adding of the C API back to the `libtensorflow.so` target in https://github.com/tensorflow/tensorflow/commit/783c52edeb3c676937dbb97ed0d40958015050d6 and https://github.com/tensorflow/tensorflow/pull/3830


adamcrume:
Thanks!

PR #59, opened 2017-02-27
Fix build script and optimize native libtensorflow.

daschl:
This changeset provides two changes to the sys build file:

 1) After 7e6ee652bac5d15b86fe521d779b9267dbecff4f (the _c) rename
    the test suite doesn't work anymore since the proper artifact
    is not linked now. This has been fixed by modifying the LIBRARY
    static to reflect the new .so file.
 2) Since the library is built to be used on this specific machine
    the --copt=-march=native flag is now passed when building TF
    to make use of all the native features possible.

@adamcrume the first change mentioned here actually makes the build work again.

### DIFF
--- a/tensorflow-sys/build.rs
+++ b/tensorflow-sys/build.rs
@@ -9,7 +9,7 @@ use std::process::Command;
 use std::{env, fs};
 use semver::Version;
 
-const LIBRARY: &'static str = "tensorflow_c";
+const LIBRARY: &'static str = "tensorflow";
 const REPOSITORY: &'static str = "https://github.com/tensorflow/tensorflow.git";
 const TARGET: &'static str = "tensorflow:libtensorflow.so";
 const TAG: &'static str = "v1.0.0";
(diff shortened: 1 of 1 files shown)


tensorflow-sys/build.rs
```diff
@@ -9,7 +9,7 @@ use std::process::Command;
 use std::{env, fs};
 use semver::Version;
 
-const LIBRARY: &'static str = "tensorflow_c";
+const LIBRARY: &'static str = "tensorflow";
```
adamcrume: Good catch.  I'm not sure why it works on my machine; I must just be "lucky".

tensorflow-sys/build.rs
```diff
@@ -79,6 +79,7 @@ fn main() {
                 .arg("build")
                 .arg(format!("--jobs={}", get!("NUM_JOBS")))
                 .arg("--compilation_mode=opt")
+                .arg("--copt=-march=native")
```
adamcrume: Nifty!

PR #61, opened 2017-02-28
`Status` refactoring.

daschl:
This changeset refactors the internals and usage of the `Status`
struct a bit to clean up the semantics. Notably:

  - It exposes the inner() as a function demanding mutable access
    since it is sharing a mutable raw pointer. Usage throghout the
    library has been fixed accordingly.
  - Docs are expanded a bit.
  - Trait impls have narrowed unsafe scope.
  - `as_result` has been renamed to `into_result` since its consuming
    self.

### DIFF
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -153,10 +153,10 @@ impl Graph {
 
     /// Returns the graph definition as a protobuf.
     pub fn graph_def(&self) -> Result<Vec<u8>> {
-        let status = Status::new();
+        let mut status = Status::new();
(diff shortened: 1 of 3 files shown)


daschl:
side note: I thought about naming the method `as_mut_ptr` since that would be the convention for a *mut T but its not exactly T :) .. so I ended up with inner(), although I would love to see a rust guideline on how to name wrapped FFI accessors.

adamcrume:
Thanks!

PR #62, opened 2017-02-28
Fix various rustc and clippy warnings.

daschl:
This changeset fixes various rustc and clippy warnings. Note
that it also gets rid of some type aliases which have been
private and are unused in the library (and since they are
private they can't be used from outside anyways).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1,5 +1,5 @@
 //! This crate provides Rust bindings for the
-//! [TensorFlow](https://www.tensorflow.org) machine learning library.
+//! [`TensorFlow`](https://www.tensorflow.org) machine learning library.
 
 #![warn(missing_copy_implementations,
         missing_debug_implementations,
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -343,6 +343,7 @@ c_enum!("Type of a single tensor element.", TF_DataType, DataType {
   /// 16-bit floating point.
   value Half = 19,
 
+  /// Tensorflow Resource (name, container, device,...)
```
adamcrume: 'f' in TensorFlow needs to be capitalized.

src/lib.rs
```diff
@@ -1003,16 +1004,13 @@ trait OperationTrait {
...
-/// TensorFlow library. TensorFlow using semantic versioning.
+/// `TensorFlow` library. `TensorFlow` is using semantic versioning.
 pub fn version() -> std::result::Result<String, Utf8Error> {
     unsafe { CStr::from_ptr(tf::TF_Version()).to_str().map(|s| s.to_string()) }
 }
 
 ////////////////////////
 
-#[deprecated(note="Use Shape instead.")]
-type TensorShape = Shape;
```
adamcrume: This was a bug and should have been public, which I fixed.  Please cancel this deletion.

daschl:
@adamcrume updated and rebased!

adamcrume:
Thanks!

PR #65, opened 2017-03-06
Install TensorFlow from a prebuilt binary when possible

adamcrume:


adamcrume:
This massively speeds up our build time, so our Travis builds don't time out any more. :smiley: Not to mention being less painful when developing locally.

jhseu:
@asimshankar fyi

daschl:
@adamcrume should we still have a switch to force build?

adamcrume:
@daschl @jhseu I thought about using a `feature`, but it doesn't seem like the right level of abstraction.  Plus, if we start configuring build details with features, we can only turn things on or off, not pass in values.  We could use an environment variable, but I'm not crazy about that, either.  What do you think?

daschl:
@adamcrume I agree, a feature is probably not the right abstraction level. For example, the openssl crate also exposes environment variables (https://github.com/sfackler/rust-openssl#manual-configuration) which I think also is the most portable.

The only thing I worry about potentially, but let me know if I'm mistaken, is that the `--copt=-march=native` compile option will be gone, right? Or are the prebuilt binaries compiled with such optimizations? On the other hand noone is preventing one from compiling manually with those flags and then using it at link time.

So bottom line is I think this is pretty great since it helps speeding up build time a lot. We should just see how we can keep the flexibility/optimization possibilities in an accessible way.

daschl:
@adamcrume the other thing which came to mind is that right now the code downloads the tarball into the target directory, so on a clean it has to be redownloaded. I wonder if we should support setting a system property for a different location so that even after a cargo clean it just picks up the tarball from that folder if it does exist?

jhseu:
@daschl Yeah, the prebuilt binaries are general and so don't include the benefits of SIMD. The performance difference can be notable (~2x). A reasonable compromise here might be to use a system-installed libtensorflow.so if it's available, or add it as an option if there's an idiomatic way to do that in Rust.

I'm not too worried about this in the long-term. When XLA is default enabled, we JIT compile with SIMD instructions.

Related: ideally TensorFlow would get statically linked in for both Rust and Go. Bazel should be getting support for building static libraries soon.

adamcrume:
I added a couple of environment variables.  What do you think?

daschl:
@adamcrume 👍

.travis.yml
```diff
@@ -1,23 +1,25 @@
...
   - source travis-ci/install.sh
 
 script:
+  - export RUST_BACKTRACE=1
   - cargo test -vv -j 2 --features tensorflow_unstable
   - cargo run --example regression
   - cargo run --features tensorflow_unstable --example expressions
   - cargo doc -vv --features tensorflow_unstable
-  - (cd tensorflow-sys && cargo test -vv -j 1)
+  - # TODO(#66): Re-enable: (cd tensorflow-sys && cargo test -vv -j 1)
```
adamcrume: Yes; this test fails on my local machine and on Travis: https://travis-ci.org/tensorflow/rust/builds/208479366

PR #68, opened 2017-03-09
Load from saved model support

Enet4:
This PR exposes TensorFlow's native capability of loading saved model bundles from a directory.

- Added TF_LoadSessionFromSavedModel function to tensorflow-sys;
- Added function `Session::from_saved_model`, which provides a safe generic API with the minimum arguments required. In the future, one might consider adding alternative functions that would let users specify `run_options` and retrieve `meta_graph_def`.

This feature will hopefully make loading pre-trained models in Rust more accessible. Please let me know if you would like a complete example or an integration test. Some of the examples found in this repository should be easily adjusted to test this feature.

Example of use:

```rust
let mut graph = Graph::new();
println!("Loading session...");
let sess = Session::from_saved_model(
    &SessionOptions::new(), &["serve"],
    &mut graph,
    "test-model").unwrap();

println!("Loaded model with {} operations.", graph.operation_iter().count());
```


src/session.rs
```diff
@@ -1,7 +1,10 @@
 use tf;
-use libc::c_int;
+use libc::{c_char, c_int};
+use std::ffi::CString;
 use std::marker;
+use std::path::Path;
 use std::ptr;
+use std::result::Result as StdResult;
```
daschl: hm I find this confusing, but if @adamcrume is good with it

src/session.rs
```diff
@@ -32,6 +35,43 @@ impl Session {
...
+        
+        let tags_cstr: Vec<_> = try!(tags.into_iter()
+                .map(|t| CString::new(t.as_ref()))
+                .collect::<StdResult<_,_>>()
+                .map_err(|_| Status::new_set(Code::InvalidArgument, "Invalid tag name").unwrap()));
+        // keeping tags_cstr to retain strings in memory
+        let tags_ptr: Vec<*const c_char> = tags_cstr.iter().map(|t| t.as_ptr()).collect();
+        
+        let inner = unsafe {
+        tf::TF_LoadSessionFromSavedModel(
```
daschl: formatting looks a bit off here, can you run it through rustfmt please?

daschl:
@Enet4 I'd love to see at least an example on this, since we are lacking them anyways and it also helps other people onboard more quickly

src/session.rs
Enet4: Well, I just needed a hint on line 56 to pluck the result out, and this approach sounded elegant to me. We could consider something else though...

Enet4:
@daschl All right, I have formatted the code (my bad!) and added an example based on regression, which was already available. Still, I'm open to file renames or other tweaks.

daschl:
@Enet4 very cool, thanks! Of course @adamcrume has the final say on this ;)

src/session.rs
```diff
@@ -32,6 +35,49 @@ impl Session {
...
+                                         .map_err(|_| {
+            Status::new_set(Code::InvalidArgument, "Invalid tag name").unwrap()
+        }));
+        // keeping tags_cstr to retain strings in memory
+        let tags_ptr: Vec<*const c_char> = tags_cstr.iter().map(|t| t.as_ptr()).collect();
+
+        let inner = unsafe {
+            tf::TF_LoadSessionFromSavedModel(options.inner,
+                                             ptr::null(),
+                                             export_dir_cstr.to_bytes_with_nul().as_ptr() as
```
adamcrume: The return value of `to_bytes_with_nul` doesn't live long enough.  You can just use `export_dir_cstr.as_ptr()`, since it already guarantees a null terminator.

Enet4:
I still wonder how I got the tag conversion right and then screwed up on the export_dir string... :thinking: Nevertheless, the changes were made. :)

adamcrume:
Thanks!

jhseu:
Note that we're preferring to add a SavedModelBundle wrapper for the return value in other languages. You need the MetaGraphDef to extract out the signatures in the SavedModel.

Enet4:
@jhseu Admittedly, I knew that at least the Java bindings would be doing it with a bundle (tensorflow/tensorflow#7134), but I found no reason to replicate that design here, at the time. But given that the meta-graph is still unreachable, I agree that we should still seek to improve this saved model API. Also, a proper MetaGraph abstraction would probably be nicer than just retrieving a byte buffer.

PR #72, opened 2017-03-17
Fix running under nightly: copy prebuilt `libtensorflow.so` to `OUT_DIR`

nbigaouette-eai:
A change in cargo nightly (merged on February 7th 2017) changed how
dynamic libraries are loaded. This prevented the `tensorflow-sys` crate's
tests to run.

To fix, simply copy the `libtensorflow.so` pre-built library to `OUT_DIR`
and change `cargo:rustc-link-search=` path so cargo can find it when
running the tests.

Note that I believe the library compiled from source does not seems to be
affected by the problem as it is already copied to `OUT_DIR`. This would
need to be verified though.

See:
* issue: https://github.com/tensorflow/rust/issues/71
* cargo's pull request: https://github.com/rust-lang/cargo/pull/3651

### DIFF
--- a/tensorflow-sys/build.rs
+++ b/tensorflow-sys/build.rs
@@ -127,11 +127,11 @@ fn install_prebuilt() {
     let output = PathBuf::from(&get!("OUT_DIR"));
     let new_library_full_path = output.join(&library_file);
     if new_library_full_path.exists() {
-        println!("File {} already exists, deleting.", new_library_full_path.display());
+        log!("File {} already exists, deleting.", new_library_full_path.display());
         std::fs::remove_file(&new_library_full_path).unwrap();
     }
 
-    println!("Copying {} to {}...", library_full_path.display(), new_library_full_path.display());
+    log!("Copying {} to {}...", library_full_path.display(), new_library_full_path.display());
     std::fs::copy(&library_full_path, &new_library_full_path).unwrap();
     println!("cargo:rustc-link-search={}", output.display());
 }
--- a/tensorflow-sys/build.rs
+++ b/tensorflow-sys/build.rs
@@ -112,7 +112,9 @@ fn install_prebuilt() {
     // Extract the tarball.
     let unpacked_dir = download_dir.join(base_name);
     let lib_dir = unpacked_dir.join("lib");
-    if !lib_dir.join(format!("lib{}.so", LIBRARY)).exists() {
+    let library_file = format!("lib{}.so", LIBRARY);
+    let library_full_path = lib_dir.join(&library_file);
+    if !library_full_path.exists() {
         extract(file_name, &unpacked_dir);
     }
 
@@ -122,7 +124,16 @@ fn install_prebuilt() {
         }); // TODO: remove
 
     println!("cargo:rustc-link-lib=dylib={}", LIBRARY);
-    println!("cargo:rustc-link-search={}", lib_dir.display());
+    let output = PathBuf::from(&get!("OUT_DIR"));
+    let new_library_full_path = output.join(&library_file);
+    if new_library_full_path.exists() {
+        println!("File {} already exists, deleting.", new_library_full_path.display());
+        std::fs::remove_file(&new_library_full_path).unwrap();
+    }
+
+    println!("Copying {} to {}...", library_full_path.display(), new_library_full_path.display());
+    std::fs::copy(&library_full_path, &new_library_full_path).unwrap();
+    println!("cargo:rustc-link-search={}", output.display());
 }
 
 fn build_from_src() {


tensorflow-sys/build.rs
```diff
@@ -122,7 +124,16 @@ fn install_prebuilt() {
         }); // TODO: remove
 
     println!("cargo:rustc-link-lib=dylib={}", LIBRARY);
-    println!("cargo:rustc-link-search={}", lib_dir.display());
+    let output = PathBuf::from(&get!("OUT_DIR"));
+    let new_library_full_path = output.join(&library_file);
+    if new_library_full_path.exists() {
```
adamcrume: I don't think deleting the old one is necessary since the copy will overwrite it anyway.

tensorflow-sys/build.rs
```diff
@@ -122,7 +124,16 @@ fn install_prebuilt() {
...
+        println!("File {} already exists, deleting.", new_library_full_path.display());
+        std::fs::remove_file(&new_library_full_path).unwrap();
+    }
+
+    println!("Copying {} to {}...", library_full_path.display(), new_library_full_path.display());
```
adamcrume: Please use the `log!` macro.

tensorflow-sys/build.rs
nbigaouette-eai: That what I though, and that's what [the documentation](https://doc.rust-lang.org/std/fs/fn.copy.html) mentions. But `unwrap()`ing the result fails. For example if I comment out the `exists()` block and simply `copy()` I get this:
```
 -> cargo test
   Compiling tensorflow-sys v0.7.0 (file://${HOME}/tensorflow_rust.git/tensorflow-sys)
error: failed to run custom build command for `tensorflow-sys v0.7.0 (file://${HOME}/tensorflow_rust.git/tensorflow-sys)`
process didn't exit successfully: `${HOME}/tensorflow_rust.git/target/debug/build/tensorflow-sys-28fb5808195bfa7e/build-script-build` (exit code: 101)
--- stdout
libtensorflow-sys/build.rs:80: binary_url = "https://storage.googleapis.com/tensorflow/libtensorflow/libtensorflow-cpu-darwin-x86_64-1.0.0.tar.gz"
libtensorflow-sys/build.rs:84: base_name = "libtensorflow-cpu-darwin-x86_64-1.0.0"
libtensorflow-sys/build.rs:93: file_name = "${HOME}/tensorflow_rust.git/tensorflow-sys/target/libtensorflow-cpu-darwin-x86_64-1.0.0.tar.gz"
libtensorflow-sys/build.rs:209: Executing "ls" "-l" "${HOME}/tensorflow_rust.git/tensorflow-sys/target/libtensorflow-cpu-darwin-x86_64-1.0.0/lib"
total 209520
-r-xr-xr-x  1 nicolas  staff  107270400 31 déc  1969 libtensorflow.so
libtensorflow-sys/build.rs:213: Command "ls" "-l" "${HOME}/tensorflow_rust.git/tensorflow-sys/target/libtensorflow-cpu-darwin-x86_64-1.0.0/lib" finished successfully
cargo:rustc-link-lib=dylib=tensorflow
libtensorflow-sys/build.rs:134: Copying ${HOME}/tensorflow_rust.git/tensorflow-sys/target/libtensorflow-cpu-darwin-x86_64-1.0.0/lib/libtensorflow.so to ${HOME}/tensorflow_rust.git/target/debug/build/tensorflow-sys-adb04dcfd9da710b/out/libtensorflow.so...

--- stderr
thread 'main' panicked at 'called `Result::unwrap()` on an `Err` value: Error { repr: Os { code: 13, message: "Permission denied" } }', /Users/rustbuild/src/rust-buildbot/slave/stable-dist-rustc-mac/build/src/libcore/result.rs:868
note: Run with `RUST_BACKTRACE=1` for a backtrace.
```

After investigation, it seems the `.so` file is `-r-xr-xr-x` (read and execute only), so the copy fails if the file already exists (it can't be overwritten as it's not writable. Making the file writable allow the copy operation to succeed.

So for the copy to succeed, I either have to chmod the `.so` file or delete the old one... Any preference?

tensorflow-sys/build.rs
nbigaouette-eai: Done in 05d4d2b9bc88e4b2793b1970857c573f4f16b8e9

tensorflow-sys/build.rs
adamcrume: Huh, I guess that makes sense.  Let's keep the delete code.  The fewer platform-dependent requirements (like chmod) that we introduce, the better.

adamcrume:
Thanks!

PR #77, opened 2017-03-20
Add bf16 type for BFloat16 tensors

adamcrume:


adamcrume:
I'm uncertain about the name, too.  It's consistent with naming for primitive types, but you could argue that it shouldn't look like a primitive type.  (num::Complex is named like a normal type, for example.)  I could rename it to BFloat16.

jhseu:
BFloat16 sounds better

Issue #69, opened 2017-03-10
Set up an automated valgrind test


adamcrume:
Not usually necessary for Rust, but since we're doing a lot of unsafe stuff with pointers, we should have a valgrind test.  Ideally, Travis would run it and fail the build if there are issues.

nbigaouette-eai:
It is always a good idea to run code/tests through valgrind. One could simply run the current tests (without docs) through valgrind, and add more that use yet uncovered code path.

Something interesting too would be to add code coverage to make sure as much as possible is covered.

And then even fuzzing; I've heard good things of `cargo fuzz`.

nbigaouette-eai:
Note that jemalloc _removed_ valgrind support (see https://github.com/jemalloc/jemalloc/commit/9a8add1510456464bc496320990ec234798bd381) but this does not seems to be included in rust's version of a jemalloc. Rust uses jemalloc @ 33184bf69813087bf1885b0993685f9d03320c69 plus a couple of patches, which is jemalloc 4.1.0 plus four patches.

Enet4:
Well, I read this recent [blog post](http://creativcoder.me/post/checking_memory_leaks_in_rust_ffi/) on using Valgrind for this very purpose, and we are still advised to change the memory allocation system (using the nightly `alloc_system` feature). Other than that, this is a useful resource to whoever is attempting to resolve this issue.

nbigaouette-eai:
Great read, thanks.

Maybe an optional combination `alloc_system`+nightly+valgrind could be added. As @adamcrume mentioned in #66, it might not be a good idea to depend on `alloc_system` and nightly for the crate, but an option to increase the testing is definitely a good idea.

adamcrume:
We might copy the suggestion at https://github.com/rust-lang/rust/issues/28224#issuecomment-281552903, which is to add a `nightly` feature and add this to the code:
```
#![cfg_attr(feature="nightly", feature(alloc_system))]
#[cfg(feature="nightly")]
extern crate alloc_system;
```
Then we can build on the nightly channel with `--features=nightly` and run valgrind against that.

nbigaouette-eai:
So I've experimented running the tests through valgrind, but it's not as trivial as it looks.

First, as mentioned above, using jemalloc (the default memory allocator in rust) will confuse valgrind. So one has to build the tests using the system allocator, which is only available when using a nightly build of rust. This means Travis would need to be setup to run nightly, possibly in addition to stable rust.

[...]

PR #78, opened 2017-03-20
Increase Travis coverage

nbigaouette-eai:
This PR adds will set Travis to follow Rust release train (stable/beta/nightly) on both Linux and OSX to increase the test platform coverage.

This should open the door to the addition of more tests, for example valgrind or sanitizers as described in #69.

Notes:
* Compilation of TensorFlow (`TF_RUST_BUILD_FROM_SRC=true`) is explicitly disabled as I have not been able to properly do so; bazel fails with `Server finished RPC without an explicit exit code` (see for example [here](https://travis-ci.org/nbigaouette-eai/rust/builds/212347865), more specifically [job 14.1](https://travis-ci.org/nbigaouette-eai/rust/jobs/212347866));
* Environment variables are moved to `env.global` instead of explicitly `export`ing them;
* Nightly rust are allowed to fail since the compiler itself can have regressions;
* The option `fast_finish` will let builds decide on success or failure even if those with `allow_failures` are not finished.
* numpy is installed using `pip` instead of `apt` as it's the recommended way of doing so.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,11 @@ env:
     - TF_RUST_BUILD_FROM_SRC=false
 
 
+matrix:
+  allow_failures:
+    - rust: nightly
+
+
 install:
   # Install dependencies only if we build tensorflow from source.
   - if [[ "$TF_RUST_BUILD_FROM_SRC" == "true" ]]; then source travis-ci/install.sh; fi
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,9 +11,14 @@ env:
     - RUST_BACKTRACE=1
     - CC="gcc-4.9"
     - CXX="g++-4.9"
+  matrix:
+    - TF_RUST_BUILD_FROM_SRC=true
+    - TF_RUST_BUILD_FROM_SRC=false
+
 
 install:
-  - source travis-ci/install.sh
+  # Install dependencies only if we build tensorflow from source.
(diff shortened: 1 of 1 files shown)


adamcrume:
Yeah, you won't be able to build from source on Travis.  One of the driving reasons for getting pre-compiled binaries into the build was to make Travis useful again.  It simply takes too long to build from source, and Travis times out.  (Or if you run with too many parallel compilations, it'll fail with `Server finished RPC without an explicit exit code` as you saw.  More than one task is too many for Travis, because the TensorFlow build is amazingly memory hungry.)

.travis.yml
```diff
@@ -5,14 +5,38 @@ dist: trusty # still in beta, but required for the prebuilt TF binaries
...
+  allow_failures:
+    - rust: nightly
+
 
 install:
-  - export CC="gcc-4.9" CXX="g++-4.9"
-  - source travis-ci/install.sh
+  # Install dependencies only if we build tensorflow from source.
+  - if [[ "$TF_RUST_BUILD_FROM_SRC" == "true" ]]; then source travis-ci/install.sh; fi
+  - if [[ "$TF_RUST_BUILD_FROM_SRC" == "true" ]]; then pip install --user numpy; fi
```
adamcrume: Good call; we were wasting build time installing things we didn't need.

adamcrume:
Thanks!

PR #82, opened 2017-05-09
Bump to tensorflow v1.1.0

nbigaouette-eai:
Bump the `tensorflow-sys` wrapper to 1.1.0.

The script `generate_bindgen_rs.sh` was used (which simply executes `bindgen --blacklist-type max_align_t target/libtensorflow-cpu-linux-x86_64-1.1.0/include/tensorflow/c/c_api.h --output src/bindgen.rs`).

Note that I also bumped the `-sys` crate version to 0.8.0 and the main one to 0.3.2. Hopefully this is the proper way?

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "tensorflow"
-version = "0.3.1"
+version = "0.3.2"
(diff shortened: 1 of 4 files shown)


adamcrume:
Thanks!

PR #84, opened 2017-05-20
Add graph def import options and clean up a few things

adamcrume:


src/expr.rs
```diff
@@ -285,20 +285,21 @@ impl<T: TensorType> ExprImpl<T> for TruncateDiv<T> {
         vec![Box::new(self.left.clone()), Box::new(self.right.clone())]
     }
 
+    // TODO: Get rid of Rc
```
adamcrume: Whoops, fixed.

src/graph.rs
```diff
@@ -283,6 +283,36 @@ impl Graph {
             status.into_result()
         }
     }
+
+    /// Import the graph serialized in `graph_def`.
+    ///
+    /// `num_return_outputs` must be the number of return outputs added (i.e. the
```
Enet4: The documentation seems to be suitable for the C API, but it's somewhat misleading in the Rust API, since `num_return_outputs` or `return_outputs` don't exist as arguments. I suppose they can just be discarded here?

adamcrume: Yep, didn't clean up after the copy/paste.  Fixed.

PR #86, opened 2017-06-04
Session runs leak memory

danieldk:
Memory leaks during repeated session runs. I noticed that the `run` method creates wrapping tensors of the input. However, the caller retains ownership of `TF_Tensor`s passed to `TF_Run` since Tensorflow 0.12.0:

https://github.com/tensorflow/tensorflow/blob/master/RELEASE.md#breaking-changes-to-the-api-2

Passing the input tensor vector from the step resolves this leak.


adamcrume:
Great catch!  We'd like to set up an automated valgrind test to catch this stuff (https://github.com/tensorflow/rust/issues/69), but it's a bit difficult.

PR #87, opened 2017-06-04
Fix a memory leak in Buffer

danieldk:
The Buffer type seems to conflate ownership of the TF_Buffer object and the
ownership of the data wrapped by the TF_Buffer object.

A TF_Buffer object created through from_ptr() is created as 'borrowed'. As
a result, the TF_Buffer object is not deallocated, leadning to a memory
leak.

This commit makes two small changes to (hopefully) remedy the leak:

- from_ptr() sets the TF_Buffer to be owned, so that the drop() will
  deallocate the TF_Buffer object. Since the TF_Buffer does not have
  a data_deallocator set, the data is not freed when Buffer is dropped.
- into_ptr() does not change the TF_Buffer ownership, but instead sets
  data_deallocator to NULL. As a consequence, the caller becomes responsible
  for freeing the data and the TF_Buffer is deallocated by drop().

### DIFF
--- a/src/buffer.rs
+++ b/src/buffer.rs
@@ -86,7 +86,7 @@ impl<T: TensorType> Buffer<T> {
         (*inner).length = len;
         Buffer {
             inner: inner,
-            owned: false,
+            owned: true,
             phantom: PhantomData,
         }
     }
@@ -96,8 +96,7 @@ impl<T: TensorType> Buffer<T> {
     /// The caller is responsible for freeing the data.
     pub unsafe fn into_ptr(mut self) -> (*mut T, usize) {
         // TODO: remove
-        // This flag is used by drop.
-        self.owned = false;
+        (*self.inner).data_deallocator = None;
         (self.data_mut(), self.length())
     }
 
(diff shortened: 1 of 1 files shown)


adamcrume:
The Buffer type is admittedly a bit of a mess.  I've thought about splitting it into a heap-allocated array type (something like Vec but fixed-length and possibly owned by C code) and a a TF_Buffer wrapper.  At the very least, tracking ownership of the array and the TF_Buffer object explicitly and separately might help.

This change seems reasonable, but I'd like to run valgrind to verify it.  There may be interesting cases like creating one with `from_c` and then calling `into_ptr`.  I'm short on time at the moment, but I'll try to get to it this evening.

danieldk:
Thanks, very much appreciated! And thanks for the binding, I really like the easy-to-follow API.

adamcrume:
I'm glad you like the API. :)

valgrind says your change improves things.  We still have some leaks, but I'll try to attack those.

adamcrume:
Thanks, by the way!

PR #89, opened 2017-06-17
Add run-valgrind

adamcrume:
This partially addresses #69, but it requires us to build TensorFlow from source, which is too slow to do on Travis.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 *~
+*.log
 Cargo.lock
 target
 travis-ci/travis_rsa
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -19,6 +19,8 @@ random = "0.12"
 
 [features]
 tensorflow_unstable = []
+# This is for testing purposes; users should not use this.
+nightly = []
 
 [workspace]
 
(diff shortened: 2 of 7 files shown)


Enet4:
It sounds like this modified version of TensorFlow is something worth caching, either [on Travis itself](https://docs.travis-ci.com/user/caching/) or somewhere else.

adamcrume:
Caching on Travis won't work, because TensorFlow takes too long to build in the first place.  Someone would have to manually build it and store it somewhere as part of upgrading the version of TensorFlow that we depend on.  @asimshankar, would we be able to use storage.googleapis.com/tensorflow for storing builds of TensorFlow with jemalloc disabled, or would we need a separate project?

asimshankar:
I don't think the TensorFlow maintainers have the bandwidth to regularly build binary artifacts both with and without jemalloc, so probably best to host the non-jemalloc artifacts somewhere else.

Though @jhseu might be in a better position to comment

adamcrume:
Sorry for the confusion -- I wasn't suggesting that the TensorFlow maintainers build binaries with jemalloc disabled; I was just asking if we could piggy-back on their storage space.

jhseu:
Sorry, we've been locking down the tensorflow GCS bucket so that only those who require access to perform releases can use it. I'm also not aware of a good alternative GCS location we have for this kind of purpose.

adamcrume:
I fixed a few bugs in the script.  I'll go ahead and merge this because a manually-run check is better than no check at all, but I'll leave #69 open.

PR #91, opened 2017-07-18
fix tensor_shape method for Graph

iduartgomez:
in the tensor_shape method for the Graph type, .len() was being called to pull the rank information from the tensor when calling C instead of .capacity(), so an error was being returned by the API (ie. _InvalidArgument: Expected rank is 0 but actual rank is 2_ on a [3, 3] tensor) as the vector was always empty (is an empty vector being passed to C so it cna be filled with the dimension information).

Fixed it and added a test to check for this in the future.

### DIFF
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -257,7 +257,7 @@ impl Graph {
             tf::TF_GraphGetTensorShape(self.gimpl.inner,
                                        output.to_c(),
                                        dims.as_mut_ptr(),
-                                       dims.len() as c_int,
+                                       n,
                                        status.inner());
             if status.is_ok() {
                 dims.set_len(n as usize);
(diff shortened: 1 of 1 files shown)


src/graph.rs
```diff
@@ -246,22 +246,26 @@ impl Graph {
...
     ///   * `output` is not in `graph`.
+    ///   * `num_dims` does not match the actual number of dimensions.
     pub fn tensor_shape(&self, output: Output) -> Result<Shape> {
         let mut status = Status::new();
         let n = try!(self.num_dims(output.clone()));
         if n == -1 {
             return Ok(Shape(None));
         }
         let mut dims = Vec::with_capacity(n as usize);
+        println!("DIMS: {}", n);
```
Enet4: You may wish to remove this line. 🙂

src/graph.rs
```diff
@@ -246,22 +246,26 @@ impl Graph {
...
         unsafe {
             tf::TF_GraphGetTensorShape(self.gimpl.inner,
                                        output.to_c(),
                                        dims.as_mut_ptr(),
-                                       dims.len() as c_int,
+                                       dims.capacity() as c_int,
```
adamcrume: Can you use `n` instead of `dims.capacity()`?  That seems less brittle.

src/graph.rs
```diff
@@ -246,22 +246,26 @@ impl Graph {
     ///
     /// Returns an error if:
     ///   * `output` is not in `graph`.
+    ///   * `num_dims` does not match the actual number of dimensions.
```
adamcrume: This should probably be removed.  If this happens, it's an internal error in our library.  I don't think users should have to worry about it.

src/graph.rs
```diff
@@ -1039,4 +1060,34 @@ mod tests {
         let status = g.import_graph_def(&[], &opts);
         assert!(status.is_ok());
     }
+
+    #[test]
+    fn test_get_tensor_shape() {
```
adamcrume: Thanks for the test!

src/graph.rs
iduartgomez: indeed, damn

src/graph.rs
iduartgomez: rustfmt :(

will make all the changes

PR #92, opened 2017-07-19
Refactor try!() into question mark operator

SirRade:
[RFC](https://github.com/rust-lang/rfcs/blob/master/text/0243-trait-based-exception-handling.md)  
TL;DR: They are equivalent with the later being the now idiomatic choice.


adamcrume:
Thanks!

PR #93, opened 2017-07-19
fix tensor_shape method for Graph

iduartgomez:
in the tensor_shape method for the Graph type, `.len()` was being called to pull the rank information from the tensor when calling C instead of _passing the rank information directly_ , so an error was being returned by the API (ie. InvalidArgument: Expected rank is 0 but actual rank is 2 on a [3, 3] tensor) as the vector was always empty (is an empty vector being passed to C so it can be filled with the dimension information).

Fixed it and added a test to check for this in the future.

**P.S:** made  a new clean pull request without all the whitespace changes

### DIFF
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -257,7 +257,7 @@ impl Graph {
             tf::TF_GraphGetTensorShape(self.gimpl.inner,
                                        output.to_c(),
                                        dims.as_mut_ptr(),
-                                       dims.len() as c_int,
+                                       n,
                                        status.inner());
             if status.is_ok() {
                 dims.set_len(n as usize);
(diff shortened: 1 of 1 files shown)


adamcrume:
Thanks!

By the way, you don't need to create a new pull request to throw out changes; you can remove unneeded changes from the latest commit by running `git revert -n HEAD && git reset -p`, unstaging the bits you want to keep, then `git commit --amend && git checkout .` and push with `-f`.

(For reference, this is an update of https://github.com/tensorflow/rust/pull/91.)

iduartgomez:
Thanks for the info, had the impression git resets were hard/impossible to pull out when you have committed to the upstream repo already. Will try it this way next time.

Thanks anyway!

sanmai-NL:
@iduartgomez: helpful note, it's the `git push -f` that's critical.

PR #94, opened 2017-07-22
'set_attr_tensor_list' accept all tensor types

iduartgomez:
No reason why only 'u8' tensors should be accepted AFAIK, fixes this.


adamcrume:
Thanks!

Issue #98, opened 2017-08-03
Add support for `TF_WhileParams`.


iduartgomez:
TF_WhileParams and related functions are the only way to build while control flow loops right now without messing without messing with protobuf defined ops.

PR #99, opened 2017-08-03
add support for `while_params`

iduartgomez:
This is a potential addition to close down #98, however is not ready for merge as there is a SIGSEGV upon calling TF_FinishWhile (freeing after failure to complete the loop, calling TF_AbortWhile, works fine though).

I have been unsuccessful trying to catch the source of the SIGSEGV so far, Valgrind runs all the tests just fine without any problems (which is misleading). Here is the partial backtrace from lldb:

```
  * frame #0: 0x00007fffefcb39bb libstdc++.so.6`std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::basic_string(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) + 11
    frame #1: 0x00007ffff4190ad3 libtensorflow.so`___lldb_unnamed_symbol450839$$libtensorflow.so + 93
    frame #2: 0x00007ffff4192be3 libtensorflow.so`___lldb_unnamed_symbol450897$$libtensorflow.so + 93
    frame #3: 0x00007ffff419219b libtensorflow.so`___lldb_unnamed_symbol450873$$libtensorflow.so + 74
    frame #4: 0x00007ffff4192249 libtensorflow.so`___lldb_unnamed_symbol450874$$libtensorflow.so + 165
    frame #5: 0x00007ffff4191aa1 libtensorflow.so`___lldb_unnamed_symbol450859$$libtensorflow.so + 165
    frame #6: 0x00007ffff4190d96 libtensorflow.so`___lldb_unnamed_symbol450845$$libtensorflow.so + 52
    frame #7: 0x00007ffff0fac3c3 libtensorflow.so`TF_AddInput + 54
    frame #8: 0x00007ffff0fb099c libtensorflow.so`___lldb_unnamed_symbol46$$libtensorflow.so + 85
    frame #9: 0x00007ffff0fb1ab4 libtensorflow.so`TF_FinishWhileHelper + 493
    frame #10: 0x00007ffff0fb247d libtensorflow.so`TF_FinishWhile + 69
    frame #11: 0x0000555555594950 tensorflow-00cb03d3251d18e8`tensorflow::graph::{{impl}}::finish(self=WhileParams @ 0x00007fffee3fdde0, graph=0x00007fffee3fe078) + 448 at graph.rs:1236
```

P.S: debugging TensorFlow, specially calling from Rust,in combination with Bazel, is painful. I've been unable to load debug symbols in lldb/gdb and debugging with assembly is not fun. Tips are welcome for this. This were my build options: `bazel build -c opt --compilation_mode dbg -c dbg --strip=never libtensorflow.so`
but nothing.

### DIFF
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -9,8 +9,10 @@ use std;
 use std::ffi::CStr;
 use std::ffi::CString;
 use std::ffi::NulError;
+use std::iter::FromIterator;
 use std::os::raw::c_void as std_c_void;
 use std::ptr;
+use std::slice;
 use std::str::Utf8Error;
 use std::sync::Arc;
 use super::buffer::Buffer;
@@ -31,13 +33,18 @@ struct GraphLifetime;
 #[derive(Debug)]
 struct GraphImpl {
     inner: *mut tf::TF_Graph,
+    is_subgraph: bool,
 }
 
 impl Drop for GraphImpl {
     /// Graph will be deleted once no more Sessions are referencing it.
     fn drop(&mut self) {
         unsafe {
-            tf::TF_DeleteGraph(self.inner);
+            // subgraphs in a while loop are destroyed when the loop is finished/aborted,
+            // this check avoids double free
+            if !self.is_subgraph {
+                tf::TF_DeleteGraph(self.inner);
+            }
         }
     }
 }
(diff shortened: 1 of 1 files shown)


adamcrume:
If you want to debug, you may want to leave off `-c opt` entirely, although as far as I know that shouldn't affect whether or not you can load debug symols.

If I run `valgrind --trace-children=yes cargo +nightly test --features='nightly tensorflow_unstable'`, it segfaults, but if I run valgrind against the test binary directly, it doesn't.  There may be a problem with the wrong libtensorflow.so being selected, or something similar.  valgrind also reports the regex library misbehaving if I run cargo with no commands, so there may be a bug upstream.  Anyway, the test fails:

    thread 'graph::tests::test_while_loop' panicked at 'assertion failed: `(left == right)` (left: `0`, right: `2`)', src/graph.rs:1377

PR #107, opened 2017-08-30
Add Session::device_list()

### LINKED ISSUE #103: Add support for DeviceList
DeviceList was added in TensorFlow 1.3.  We need to add support for it.

adamcrume:
Fixes #103

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -1,5 +1,6 @@
 use tf;
 use libc::{c_char, c_int};
+use std::ffi::CStr;
 use std::ffi::CString;
 use std::marker;
 use std::path::Path;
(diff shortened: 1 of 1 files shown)


jhseu:
@saeta fyi

saeta:
LGTM; not for this PR, but I wonder if we can augment Status to support the `?` operator...

adamcrume:
@saeta I was wondering the same thing.  There is an experimental `std::ops::Try` trait, and there's also `Status::into_result()` (which could be changed to an `impl Into<Result<(), Status>> for Status`, but I think I wasn't sure about making it part of the public API).  A problem with both is that they consume the Status, so we would need to allocate a new `TF_Status` for every function call; we couldn't reuse them without gymnastics.  Probably not the end of the world, but I'd prefer to avoid allocations if they're easy to avoid.  Maybe the way to go is add a `try_status!` macro.

PR #110, opened 2017-10-25
Add builder method to set tensor values.

iduartgomez:
Allows for ergonomic construction of tensors like:

`let a = Tensor::new(&[2, 2]).with_values(&[0_i32, 1, 2, 3]);`

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -835,6 +835,24 @@ impl<T: TensorType> Tensor<T> {
         }
     }
 
+    /// Sets (copies) the tensor values to the provided ones.
+    /// 
+    /// ```
+    /// # use tensorflow::Tensor;
+    /// let a = Tensor::new(&[2, 2]).with_values(&[0_i32, 1, 2, 3]).unwrap();
+    /// ```
+    pub fn with_values(mut self, value: &[T]) -> Result<Self> {
+        if self.len() != value.len() { 
+            return Err(invalid_arg!(
+                "length of values array ({}) is not equal to tensor total elements ({})", 
+                value.len(), self.len()));
+        }
+        for (e, v) in self.iter_mut().zip(value) {
+            e.clone_from(v);
+        }
+        Ok(self)
+    }
+
     /// Returns the tensor's dimensions.
     pub fn dims(&self) -> &[u64] {
         &self.dims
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -835,6 +835,27 @@ impl<T: TensorType> Tensor<T> {
...
+    /// If `values.len() != self.len()`.
+    pub fn with_values(mut self, values: &[T]) -> Self {
+        assert!(
+            self.len() == values.len(), 
+            "length of values array ({}) is not equal to tensor total elements ({})", 
+            values.len(), self.len()
+        );
+        for i in 0..self.len() {
+            self[i] = values[i].clone();
+        }
```
Enet4: We can make this loop more idiomatic. Can you try this instead?

```rust
for (e, v) in self.iter_mut().zip(values) {
    *e = v.clone();
}
```

iduartgomez:
Sure, done! Also changed the "From" trait implementation to be more idiomatic (just copied that).

src/lib.rs
sanmai-NL: What about [`clone_from`](https://doc.rust-lang.org/std/clone/trait.Clone.html#method.clone_from) instead?

Enet4: @sanmai-NL Yes, good thinking. `e.clone_from(v)` should work here as well.

src/lib.rs
```diff
@@ -835,6 +835,27 @@ impl<T: TensorType> Tensor<T> {
...
+    /// 
+    /// ```
+    /// # use tensorflow::Tensor;
+    /// let a = Tensor::new(&[2, 2]).with_values(&[0_i32, 1, 2, 3]);
+    /// ```
+    ///
+    /// # Panic
+    /// If `value.len() != self.len()`.
+    pub fn with_values(mut self, value: &[T]) -> Self {
+        assert!(
```
adamcrume: Bad user input shouldn't produce panics (if we can avoid it).  This should return a `Result`, instead.  You could use `invalid_arg!` here.  Callers can always call `unwrap()` if they're sure they did the right thing.

iduartgomez:
Done.

sanmai-NL:
@iduartgomez: https://github.com/tensorflow/rust/pull/110#discussion_r146876664

iduartgomez:
All changes have been done, should be fine now.

adamcrume:
Thanks!

PR #113, opened 2017-11-18
Tensor: data for C types is not initialized to zero.

danieldk:
`Tensor::new` does not zero-initialize tensors of native C types.

This resulted in some interesting non-deterministic errors ;).

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -810,6 +810,12 @@ impl<T: TensorType> Tensor<T> {
                     dims.len() as c_int,
                     total * mem::size_of::<T>(),
                 );
+
+                // Zero-initialize allocated memory.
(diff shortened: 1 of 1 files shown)


crackcomm:
Maybe function zeros like in numpy would be better to avoid unnecessary writes on new.

adamcrume:
@danieldk Great catch, and thanks!

@crackcomm `Tensor::new` is intended to behave like `np.zeros`.  The failure to initialize the memory is a regression.  If we want a function to return an uninitialized Tensor, it should be clearly named (like `Tensor::uninitialized` or `Tensor::new_uninitialized`) and would have to be `unsafe`.

crackcomm:
@adamcrume thanks

PR #100, opened 2017-08-11
A few tweaks to allow building on Windows

Tomcc:
I wanted to get this to work on Windows because it's my preferred dev environment.

Building Tensorflow from source on Windows is not supported and extremely difficult, but fortunately it's still possible to steal the binaries from the official `pip` installation and link them with this wrapper (with enough magic).

This PR lets you define a `TF_RUST_PREBUILT_DIR` env variable that points to a folder containing `tensorflow.dll` and `tensorflow.lib` that will be used by `build.rs` to link. This could possibly be useful on other platforms as well if you don't want to `install` the binaries you built yourself.

I've also changed `buffer.rs` to require Nightly (but only on Windows) in order to replace `posix_memalign`. 

From my tests with the examples, everything seems to work alright. I don't mind if you prefer not to integrate this, but I thought it could be useful to someone else!
It's not particularly clean but it makes setting up on Windows at least possible.

PR #114, opened 2017-11-28
Tweaks for Windows

bekker:
I made some tweaks for Windows, without knowing that #100 already exists. It seems to take a while until that pr gets merged, so I open another.

Used lib file generated with https://github.com/bekker/tensorflow-windows-lib.

Basically same as #100, except:
1) For `posix_memalign`, use [external crate](https://github.com/jonas-schievink/aligned_alloc.rs), not nightly features. Therefore works with the stable rust. The crate is essentially same for Posix, but some evil tweaks for Windows.

2) Find `tensorflow.lib` from `PATH` environment variable and use it, rather than defining custom env var.

I wrote a simple mnist model with CNN and it worked well with cuda enabled.

### DIFF
--- a/tensorflow-sys/build.rs
+++ b/tensorflow-sys/build.rs
@@ -20,6 +20,7 @@ use tar::Archive;
 
 const FRAMEWORK_LIBRARY: &'static str = "tensorflow_framework";
 const LIBRARY: &'static str = "tensorflow";
+const WINDOWS_LIB: &'static str = "tensorflow.lib";
 const REPOSITORY: &'static str = "https://github.com/tensorflow/tensorflow.git";
 const FRAMEWORK_TARGET: &'static str = "tensorflow:libtensorflow_framework.so";
 const TARGET: &'static str = "tensorflow:libtensorflow.so";
@@ -38,6 +39,11 @@ macro_rules! log {
 macro_rules! log_var(($var:ident) => (log!(concat!(stringify!($var), " = {:?}"), $var)));
 
 fn main() {
+    if check_windows_lib() {
+        log!("Returning early because {} was already found", WINDOWS_LIB);
+        return;
+    }
+
     if pkg_config::find_library(LIBRARY).is_ok() {
         log!("Returning early because {} was already found", LIBRARY);
         return;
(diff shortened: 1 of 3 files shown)


tensorflow-sys/build.rs
```diff
@@ -47,13 +53,34 @@ fn main() {
...
+    false
+}
+
+#[cfg(target_env = "msvc")]
+fn check_windows_lib() -> bool {
+    if let Ok(path) = env::var("PATH") {
+        for p in path.split(";") {
+            let path = Path::new(p).join(WINDOWS_LIB);
+            if path.exists() {
+                println!("cargo:rustc-link-lib=dylib={}", LIBRARY);
```
adamcrume: As of TensorFlow 1.4, we also need to link in the `tensorflow_framework` library.

bekker: I have no knowledge on `tensorflow_framework`, but I thought it'd be unnecessary for C API which rust bindings depend on.

> We'd like the C/C++ APIs (libtensorflow.so/libtensorflow_cc.so) to be usable on their own, and those have a linker script so that they don't expose extra symbols (preventing issues like the one you've reported). https://github.com/tensorflow/tensorflow/issues/14267

Does above mean that C API would 'never' use `tensorflow_framework`, or 'optionally'?

If it's optional, I think we should add `framework` feature once framework functionality  implemented on this crate(Is it already done?), and let user choose whether to include `framework` or not. Until that, only with `tensorflow` library seems to be OK.

Also I've noticed pkg-config for `tensorflow` does not include `tensorflow-framework`. Maybe will be included in the future.

bekker:
@adamcrume I couldn't find separate binary file for `tensorflow_framework` from Windows python wheels. Current CMake builds, which is official Windows build right now, don't seem to build separate `framework` binary. I was looking into https://github.com/tensorflow/tensorflow commits regarding separate `framework` shared object, but seems no one touched CMake. Could you check?

And even for linux, `libtensorflow_framework.so` is necessary for `libtensorflow.so` but it seems linking against it should be avoided, according to tensorflow/tensorflow#14267. Would it be OK to remove the `libtensorflow_framework.so` linking from `tensorflow-sys/build.rs`?

adamcrume:
@jhseu Do you know if the Windows build separates the tensorflow and tensorflow_framework libraries, or do you know who would know?

jhseu:
Yeah, I think we don't separate it out for Windows, not for any good reason, but just because no one has done it yet. @allenlavoie would know best, though.

allenlavoie:
The TF Windows builds are both monolithic (no libtensorflow_framework). The CMake build because it hasn't been changed (there are no plans to do that AFAIK), and the Bazel build because this kind of dynamic linking doesn't work yet (https://github.com/tensorflow/tensorflow/pull/13240).

adamcrume:
Sounds like we're good to go, so I'll go ahead and merge.  I don't have a Windows environment, so I appreciate development and testing for Windows.  Thanks!

PR #117, opened 2017-12-24
Update documentation on SessionOptions.set_config.

thedodd:


src/lib.rs
```diff
@@ -842,15 +842,15 @@ impl<T: TensorType> Tensor<T> {
...
+    ///
     /// ```
     /// # use tensorflow::Tensor;
     /// let a = Tensor::new(&[2, 2]).with_values(&[0_i32, 1, 2, 3]).unwrap();
     /// ```
     pub fn with_values(mut self, value: &[T]) -> Result<Self> {
-        if self.len() != value.len() { 
+        if self.len() != value.len() {
             return Err(invalid_arg!(
-                "length of values array ({}) is not equal to tensor total elements ({})",
```
thedodd: The trimmed trailing white-space was simply due to editor save. Seems to be solid though.

src/lib.rs
```diff
@@ -485,8 +485,8 @@ impl SessionOptions {
 
     /// Set the config.
     ///
-    /// `config` should be a serialized brain.ConfigProto proto.
-    /// Returns an error if config was not parsed successfully as a ConfigProto.
+    /// `config` should be a serialized [`ConfigProto` proto](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/protobuf/config.proto).
+    /// Returns an error if config was not parsed successfully as a `ConfigProto`.
```
thedodd: Seemed like a good idea to actually go ahead and get that link in place as I had to go looking for that information. Let me know if you thinking it would be best to pin to a specific release version as opposed to just master.

adamcrume: master is fine.  If we pinned it to a specific version, it'd fall out of sync with the version we pull in.

adamcrume:
Thanks!

PR #123, opened 2018-01-25
Thread-safe tensors for C-compatible types

danieldk:
We would like to use a Tensorflow model in a multi-threaded server application. This should not be a problem, since Tensorflow is thread-safe (except for some session operations). However, the Rust compiler cannot derive `Send`/`Sync`. In most cases this is caused by the use of raw pointers and easy to solve by explicitly implementing these traits.

One exception is the Tensor data structure, since it uses `Cell` interior mutability for (lazily) unpacking tensors for types where the Rust and C representations are not identical.

I first considered replacing these cells by `RwLock` or `Mutex`. However, this would pollute the interface with lock guards.

Instead, I opted for another approach, implementing two different data structures:

1. `TensorDataCRepr` for types with identical C/Rust representations.. This data structure does not use interior mutability.
2. `TensorDataNoCRepr` for types with different C/Rust representations. This continues to use interior mutability.

The `Tensor` struct then uses one of these data structures based on the type parameter. In this approach the interface of `Tensor` does not change at all.

`TensorDataCRepr` is then marked as `Send`/`Sync` such that Tensors of C-compatiblity types can be used across threads.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -868,13 +868,6 @@ where
     }
 }
 
-impl<T> TensorDataCRepr<T>
-where
-    T: TensorType,
-{
-    fn drop_tensor(&mut self) {}
-}
-
 impl<T: TensorType> Deref for TensorDataCRepr<T> {
     type Target = [T];
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -805,6 +805,9 @@ where
     phantom: PhantomData<T>,
 }
 
+unsafe impl<T> Send for TensorDataCRepr<T> where T: TensorType {}
+unsafe impl<T> Sync for TensorDataCRepr<T> where T: TensorType {}
+
 impl<T: TensorType> Drop for TensorDataCRepr<T> {
     fn drop(&mut self) {
         if !self.inner.is_null() {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -866,7 +866,7 @@ where
         })
     }
 
-    fn as_mut_ptr(&self, dims: &Vec<u64>) -> Result<*mut tf::TF_Tensor> {
+    fn as_mut_ptr(&self, _dims: &Vec<u64>) -> Result<*mut tf::TF_Tensor> {
         assert!(!self.inner.is_null());
         Ok(self.inner)
     }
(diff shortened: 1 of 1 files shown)


src/lib.rs
```diff
@@ -921,34 +1014,103 @@ impl<T: TensorType> Tensor<T> {
...
+    }
+
+    /// Returns the tensor's dimensions.
+    pub fn dims(&self) -> &[u64] {
+        &self.dims
+    }
+
+    // Wraps a TF_Tensor. Returns None if types don't match.
+    unsafe fn from_tf_tensor(tensor: *mut tf::TF_Tensor) -> Option<Self> {
+        if DataType::from_c(tf::TF_TensorType(tensor)) != T::data_type() {
```
adamcrume: Since this check is done inside `T::InnerType::from_tf_tensor`, it doesn't need to be kept here.

src/lib.rs
```diff
@@ -921,34 +1014,103 @@ impl<T: TensorType> Tensor<T> {
...
+        self.unpack();
+        unsafe { slice::from_raw_parts(self.data.get(), self.data_count) }
+    }
+}
+
+impl<T: TensorType> DerefMut for TensorDataNoCRepr<T> {
+    #[inline]
+    fn deref_mut(&mut self) -> &mut [T] {
+        self.unpack();
+        if !T::is_repr_c() {
```
adamcrume: `is_repr_c()` is guaranteed to be false here.

adamcrume:
Thanks!

By the way, I hadn't really considered the multithreaded use case for the client, since I assumed most of the heavy lifting would be within TensorFlow itself.  Can you elaborate on your use case?

danieldk:
I use Tensorflow in my dependency parser. I was fine with single-threaded use, since I am just parsing corpora from files. When I parse large corpora, I will use Tensorflow's parallelism and process parallelism (`xargs -P`) to parse on a many-core machine.

However, some colleagues would like to use my parser in a web service. I considered several options:

* A single-threaded server. This has the disadvantage that small (interactive) requests can be blocked by large requests.
* Using a thread pool with a couple of thread-local models. But it seems to be wasteful to load the graph and embedding matrices a couple of times.
* A multi-threaded server with one thread-local model where instances from different threads are batched together or batches from different threads are interleaved and then processed by the thread-local model. This would probably have worked ok, but would complicate the logic in the server considerably.

So, the easiest solution seems to be to create one graph and let the threads create/run their own sessions.

adamcrume:
Ah, of course.  I've been focused on the training side of things and not thinking much about the serving side.

PR #125, opened 2018-01-30
Implement Send + Sync for GraphImpl/Operation.

danieldk:
According to the Tensorflow C API documentation:

    Graphs may be shared between sessions. Graphs are thread-safe when
    used as directed below.

Which presumably means when you use the C API as intended --- the C API uses a mutex to control graph access/modification.

With this change, it is possible to share a graph between sessions in different threads.


adamcrume:
Thanks!

PR #132, opened 2018-02-17
Upgrade to TensorFlow 1.6.0-rc1

adamcrume:


adamcrume:
@jhseu It looks like the TensorFlow 1.5 prebuilt binaries were linked against a newer libstdc++, and I can't for the life of me figure out how to get Travis to cooperate and link against a new enough version, even with the newer version installed.  There's so many layers of indirection, with cargo calling rustc which calls cc which calls the linker.  Any thoughts?

jhseu:
Can you try TensorFlow 1.6rc1? We had a temporary attempted upgrade to Ubuntu 16.04 with TF 1.5, but reverted back to Ubuntu 14.04. It's possible that libtensorflow wasn't updated after that happened.

Ubuntu 16.04 required a new libstdc++ and glibc, which is why we reverted it.

adamcrume:
The 1.6.0-rc1 build seems to work fine.  I guess I can go ahead an upgrade to that, and I'll just have to remember to upgrade to a non-RC build before releasing again.  By the way, Travis doesn't support Ubuntu newer than 14.04, unfortunately.

PR #159, opened 2018-04-17
Add an example for saving and loading checkpoints

bekker:
- Some of the models being provided as pb and some being not is weird. Provide all pbs, so that you can run examples without Python.
- Renamed directories for saved model pbs and saved data, for consistency.
- Added an example for saving and loading checkpoints.


adamcrume:
Looks good, and thanks for the cleanup!

PR #160, opened 2018-04-22
Add support for while loops

adamcrume:
Adds WhileParams for creating while loops.

@jhseu Can you double-check that the public API for this makes sense?  I thought about creating a `while_loop` function, but it looks like options might be added in the future, so I thought a builder made more sense.

### DIFF
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -333,6 +333,32 @@ impl Graph {
         }
     }
 
+    /// Finds a unique operation name.  The pattern must contain exactly one
+    /// '{}' placeholder to indicate where a unique ID can be inserted, e.g.
+    /// 'Add_{}' or 'while_loop_{}/Merge', and the function returns an integer
+    /// which, when inserted into the placeholder, yields an operation name
+    /// which does not appear in the graph.
+    pub(crate) fn generate_operation_name(&self, operation_name_pattern: &str) -> Result<i64> {
+        let parts: Vec<_> = operation_name_pattern.split("{}").collect();
+        if parts.len() != 2 {
+            return Err(invalid_arg!(
+                "operation_name_pattern must contain placeholder"
+            ));
+        }
+        // Can't use format! because its argument must be a string literal.
+        let mut i = 0;
+        loop {
+            let name = format!("{}{}{}", parts[0], i, parts[1]);
+            let c_name = CString::new(name)?;
+            unsafe {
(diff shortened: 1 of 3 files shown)


adamcrume:
Rust doesn't let you use a struct literal unless you specify a value for every field, so we wouldn't want to let users specify an options struct that way, because many users would specify None, None, None, etc., and it would break users when we add options.  Using the options struct would then have to look like

```rust
let opts = WhileOptions::new();
opts.name = Some(name);
let my_loop = while_loop(graph, cond, body, inputs, &opts)?;
```

Actually, in light of the [Rust style guide section on builders](https://doc.rust-lang.org/1.0.0/style/ownership/builders.html), if we stick with the builder, I might want to change it from the current

```rust
let mut params = WhileParams::new(graph, cond, body, inputs)?;
params.set_name(name)?;
let my_loop = params.finish()?:
```

to

```rust
let my_loop = WhileBuilder::new(graph, cond, body, inputs)?.name(name)?.finish()?;
```

If we went with a single function instead, it would be

```rust
let my_loop = while_loop(graph, cond, body, inputs, None, 10, True, False, Some(name), None)?;
```

and would break if options were ever added.  (In addition to all struct fields being required, all function parameters are required, too.  Rust doesn't support default parameters or overloading.)

I don't have any solid references to back me up, but I think builders are more idiomatic in Rust than options structs.

What do you think?

Enet4:
I might as well add that, although it is not applicable here, there is a syntax for expanding a base struct value with specific fields, which would allow one to write optional values without expanding `None`s everywhere:

```
let foo = Foo {
    x: Some(5),
    y: Some(2),
    .. Default::default()
};
```

But in this case (from what I gather), the struct should not implement `Default`, and we don't have any other base struct to construct new instances from.

I agree that the builder pattern is idiomatic here, and whether they are consuming or non-consuming, should be chainable. What I don't see very often is the field assignment method returning a `Result<Builder>`, but I suppose this is justified here. The alternative would be delegating all conversions to `finish`, which in turn would go a bit against the fail fast principle.

jhseu:
Yeah, a Builder makes sense.

More generally, most of the language bindings have some form of options because ops often specify a default for attributes. If you have higher-level wrappers for ops, it might make sense to use the same idiom for while loops and op creation.

A more egregious example is here:
https://www.tensorflow.org/api_docs/python/tf/layers/batch_normalization

adamcrume:
Let's use this API style and see how it works out.

Issue #162, opened 2018-05-13
Contributing TFRecordWriter and TFExampleParser


masonk:
In my own project, I am using Rust to write TFRecords full of TFExamples to disk.

The code for doing this is something that might be useful to many projects that import the `tensorflow` crate, and I'd like to submit a PR for these utils.

However, it would be nice to have an agreement from package maintainers on the specifics of the design _before_ I code up the CL. To give an example of outstanding questions:

1. Do you even want `RecordWriter` and/or `ExampleWriter` in this crate?
2. Namespace? By analogy to the Python API, what about `tensorflow::io`?
3. File structure? `src/io.rs`?

adamcrume:
I like the idea, and I appreciate asking for direction before submitting a pull request.  I have a few questions; please don't be scared off.  :)

In response to your questions:
1. In a general sense, yes.  I think if something shows up in the standard TensorFlow API, it probably makes sense to have it in this crate.  We use the C API as much as possible, though, and this isn't currently exposed through that API, so I assume this would be pure Rust?  If so, it would have to come with the caveat that if this functionality is exposed through the C API later, we may switch to use that in order to ensure consistent behavior with other language bindings.
2. That sounds fine.  We do need to start organizing our namespace, which is a mess at the moment.
3. `src/io.rs` sounds good to me.

Would this introduce additional dependencies?  If this requires a protobuf implementation, we would want to be careful to choose the right crate.  Likewise if there are any file format dependencies.

Does this include something like an `Example` struct?  What would that look like?  We currently can't express something like `Map<String, Tensor>` because `Tensor` is generic in the element type, so you couldn't have a map containing both int and float tensors, for example.  We could expose something similar to the internal `AnyTensor` trait, and then store a `Map<String, Box<AnyTensor>>`.

@jhseu Do you know if record/example IO will ever be provided via the C API?

masonk:
Yes, RecordWriter is pure Rust. Its deps are `crc32c` and `byteorder`. I have no qualms with replacing the code with externs to C, of course.  I'm pretty sure the binary format of a TFRecord is stable at this point, so I think the best reason to do it would be perf.

As for creating Examples, mine are serialized using the `protobuf` crate. I haven't spent any time wrapping the raw protobuf gencode API . I will put some put thought into an Example API, as there are a few layers of boilerplate which kind of clutter things. And, yes, this way would entail a `protobuf` dep. I think it would make sense to make this an optional dep.

[...]

PR #163, opened 2018-05-20
tensorflow::io::RecordWriter

masonk:
As discussed in issue #162, here is a pure-Rust RecordWriter.


masonk:
Looks like `crc32c` has a feature gate which breaks the stable builds of Rust. Will look for alternatives later.

Cargo.toml
```diff
@@ -14,6 +14,8 @@ libc = "0.2"
 aligned_alloc = "0.1"
 num-complex = { version = "0.1.40", default-features = false }
 tensorflow-sys = { version = "0.12.0", path = "tensorflow-sys" }
+byteorder = "*"
```
masonk: I noticed you have Cargo.lock in .gitignore; would you like me to reify the version number here instead?

(Also, out of curiosity, why not use the Cargo.lock mechanism?)

Cargo.toml
```diff
@@ -14,6 +14,8 @@ libc = "0.2"
...
+crc32c = "*"
```
masonk: I honestly can't find a stable crc32c crate (everyone's using feature gates to get assembly instructions). I'll probably do one myself.

Cargo.toml
Enet4: This crate is a library, and in accordance to the [Cargo docs](https://doc.rust-lang.org/cargo/guide/cargo-toml-vs-cargo-lock.html), libraries don't have a Cargo.lock. I would strongly recommend these new dependencies to have a more appropriate version range.

adamcrume: We definitely need minimum versions for the dependencies.

Cargo.toml
adamcrume: What about the `crc` crate?  It has a thousand times as many downloads as `crc32c` and doesn't seem to require anything but the stable channel.  I don't mind requiring a different release channel if it's feature-gated, but I think we've been doing well to stick to the stable channel for default builds.

Cargo.toml
masonk: done

src/io.rs
```diff
@@ -58,6 +57,63 @@ impl<W> RecordWriter<W> where W: Write {
     }
 }
 
+// this pure-rust Castagnoli CRC32 algorithm is lifted from 
+// rust-snappy (https://github.com/BurntSushi/rust-snappy/blob/master/src/crc32.rs)
```
adamcrume: I can't allow code that you didn't write, because of Google's CLA.  There are also license concerns.

Cargo.toml
masonk: I'll take a look.

src/io.rs
masonk: I'll take a look at crc32

src/io.rs
```diff
@@ -0,0 +1,94 @@
+//! A module for writing TFRecords, Tensorflow's preferred on-disk data format.
+//!
+//! See the [tensorflow docs](https://www.tensorflow.org/api_guides/python/python_io#tfrecords-format-details) for details of this format.
+
+extern crate byteorder;
+extern crate crc32c;
+
+use std::io;
+use std::io::prelude::*;
```
adamcrume: Please don't use wildcards in imports.  It makes it difficult to tell where imported symbols came from and which traits are in use.

src/io.rs
```diff
@@ -0,0 +1,94 @@
...
+
+        uint64 length
+        uint32 masked_crc32_of_length
+        byte   data[length]
+        uint32 masked_crc32_of_data
+        and the records are concatenated together to produce the file. CRCs are described here [1], and the mask of a CRC is
+        [1] https://en.wikipedia.org/wiki/Cyclic_redundancy_check
+        masked_crc = ((crc >> 15) | (crc << 17)) + 0xa282ead8ul
+        */
+        let mut len_bytes = vec![];
```
adamcrume: Could you change `len_bytes`, `len_crc32c_bytes`, and `bytes_cr32_bytes` from vectors to arrays; i.e. `[u8; 8]` and `[u8; 4]`?  That would avoid heap allocations, since they're small and fixed-size.

src/io.rs
masonk: Done.

src/io.rs
masonk: Good catch, thank you. Done.

adamcrume:
Thanks for the contribution!

PR #165, opened 2018-06-07
Rename StepWithGraph to SessionRunArgs and improve documentation

### LINKED ISSUE #164: Documentation for when/how to use "Step"
The C++ and Go SDKs don't have a concept of "Step", to run a graph.  Instead, they specify inputs and outputs , like:

```cpp
  std::vector<Tensor> outputs;
  ClientSession session(root);
  // Run and fetch v
  TF_CHECK_OK(session.Run({v}, &outputs));
  // Expect outputs[0] == [19; -3]
  LOG(INFO) << outputs[0].matrix<float>();
  return 0;
```

It would be great for someone like me who is coming from those SDKs what the concept behind "Step" is in the Rust SDK:

```rust
    // Run the Step
    let mut step = StepWithGraph::new();
    step.add_input(&graph.operation_by_name_required("x")?, 0, &x);
    step.add_input(&graph.operation_by_name_required("y")?, 0, &y);
    let z = step.request_output(&graph.operation_by_name_required("z")?, 0);
    session.run(&mut step)?;
```

Thank you!

adamcrume:
Fixes #164

### DIFF
--- a/examples/addition.rs
+++ b/examples/addition.rs
@@ -14,8 +14,8 @@ use tensorflow::Graph;
 use tensorflow::ImportGraphDefOptions;
 use tensorflow::Session;
 use tensorflow::SessionOptions;
+use tensorflow::SessionRunArgs;
 use tensorflow::Status;
-use tensorflow::StepWithGraph;
 use tensorflow::Tensor;
 
 fn main() {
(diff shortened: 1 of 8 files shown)


examples/addition.rs
```diff
@@ -54,15 +54,15 @@ fn run() -> Result<(), Box<Error>> {
...
 
-    // Run the Step
-    let mut step = StepWithGraph::new();
-    step.add_input(&graph.operation_by_name_required("x")?, 0, &x);
-    step.add_input(&graph.operation_by_name_required("y")?, 0, &y);
-    let z = step.request_output(&graph.operation_by_name_required("z")?, 0);
-    session.run(&mut step)?;
+    // Run the graph.
+    let mut args = SessionRunArgs::new();
+    args.add_input(&graph.operation_by_name_required("x")?, 0, &x);
```
jhseu: (optional) inputs/outputs are named feeds and fetches in TensorFlow nomenclature as related to session runs. Either way works, though.

jhseu: Yeah, perhaps rename them. Up to you.

PR #171, opened 2018-08-24
Use TensorType's datatype instead of hardcoded DataType::Float

regiontog:
Using ```Expr<f64>``` placeholders returns ```Err``` when attempting to run the session with f64 values.


adamcrume:
Thanks!

PR #172, opened 2018-08-25
In expr::Variable and expr::Placeholder set shape attr when creating their operations

regiontog:


src/expr.rs
```diff
@@ -504,6 +588,25 @@ impl<T: TensorType> Assign<T> {
...
+    pub fn to(variable: Expr<T>, iterable: impl Iterator<Item = T>) -> Expr<T> {
+        let constant = if let ShapeHint::Exactly(shape) = variable.expr.shape_hint() {
+            let values: Vec<_> = iterable
+                .take(shape.iter().fold(1, |size, dim| size * dim) as usize)
+                .collect();
+
+            Constant::new_expr(
+                Tensor::new(shape)
+                    .with_values(&values)
+                    .expect("iterable to contain enough values to fill variable"),
```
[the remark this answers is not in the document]
regiontog: I agree, it does make the call site a bit weird as you have to handle two results (compile and Assign::at). Would you prefer to delay error propagation until create_operation or fail early?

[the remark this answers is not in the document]
adamcrume: I'd slightly prefer to fail early.  Trying to delay creating the Constant complicates things like the Display impl and children().

src/lib.rs
```diff
@@ -1493,4 +1553,24 @@ mod tests {
...
+    #[test]
+    fn tensor_display() {
+        let tests = [
+                ("1", &[][..], &[1][..]),
+                ("[1]", &[1], &[1]),
+                ("[1, 2]", &[2], &[1, 2]),
+                ("[[1, 2], [3, 4]]", &[2, 2], &[1, 2, 3, 4]),
+                ("[[[1], [2]], [[3], [4]]]", &[2, 2, 1], &[1, 2, 3, 4]),
+                ("[[], []]", &[2, 0], &[]),
+                ("[[], []]", &[2, 0, 2], &[]),
```
regiontog: Is this is the least surprising format for these edge case shapes?

src/expr.rs
```diff
@@ -504,6 +595,24 @@ impl<T: TensorType> Assign<T> {
...
+            let values: Vec<_> = iterable
+                .take(shape.iter().product::<u64>() as usize)
+                .collect();
+
+            Constant::new_expr(
+                Tensor::new(shape)
+                    .with_values(&values)?
+            )
+        } else {
+            panic!("Cannot assign to expression with unknown size!")
```
regiontog: Are we okay with panic'ing here? Is trying to assign to something that isn't variable something a user of the library would want to handle at runtime?

src/lib.rs
adamcrume: I think so.  If you go to play.rust-lang.org and enter:
```rust
fn main() {
  println!("{:?}", <[[[i32; 2]; 0]; 2]>::default());
}
```
you get `[[], []]`, and it would probably be least surprising to be consistent with how arrays are formatted.

src/expr.rs
adamcrume: If it's a condition that a user of the library can trigger, it should be an error, not a panic.

src/expr.rs
```diff
@@ -77,6 +77,15 @@ impl<T: TensorType> From<T> for Expr<T> {
 
 ////////////////////////
 
+/// Enum of an expr's possible shape states
+
+pub enum ShapeHint<'a> {
```
adamcrume: Please add a #[derive(Debug)] and variant-level docs.

src/lib.rs
```diff
@@ -1213,6 +1213,66 @@ impl<T: TensorType + PartialEq> PartialEq for Tensor<T> {
     }
 }
 
+fn write_tensor_recursive<T: Display + Debug>(f: &mut Formatter, shape: &[u64], values: &[T]) -> ::std::fmt::Result {
```
adamcrume: This doesn't need the bound on Debug.

src/lib.rs
```diff
@@ -1213,6 +1213,66 @@ impl<T: TensorType + PartialEq> PartialEq for Tensor<T> {
...
+    match shape.len() {
+        // Handle special case of a scalar tensor.
+        0 => write!(f, "{}", values[0]),
+        1 => {
```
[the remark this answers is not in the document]
regiontog: Yeah, I put it in because I felt like it was unnecessary to split the values into a bunch of arrays with only 1 element and then do a function call. But the implementation is much nicer without. I guess the cost is at least one function call per value now, unless the function is inlined by rust.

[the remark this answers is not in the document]
adamcrume: This new version may be less efficient, but it's more maintainable and less likely to contain bugs.  If someone runs into a situation where displaying tensors is a major performance cost, we can revisit this, but I suspect that in real-world situations, the cost of formatting tensors should be negligible compared to the rest of the program.

adamcrume:
Thanks for sticking through the thorough review!

PR #173, opened 2018-08-25
Implement public function Expr::new so that external projects can have custom operations.

regiontog:
I'm trying to create a project with a function that returns an ```Expr```. The motivation for this is so that I can continue to apply the operations (+, -, *, /) provided by the ```expr``` module. However the ```expr``` field is private so there is no way that I can see to create a ```Expr``` struct in a project external to tensorflow. Implementing ```ExprImpl``` is also difficult as ```Expr``` does not expose e.g. ```op_level```.

Creating a public ```Expr::new``` function that creates an ```Expr``` from an ```ExprImpl``` and implementing ```ExprImpl<T>``` for ```Expr<T>``` seems to be sufficient for my needs, however I have yet to implement the ```derivative_for_var``` properly. I'm also unsure if you would need more ```OpLevel``` variants or if the ones currently available covers all levels needed for any mathematical operation.

Is this something you would like to support?

### DIFF
--- a/src/expr.rs
+++ b/src/expr.rs
@@ -51,13 +51,40 @@ pub enum OpLevel {
 ///
 /// This is separate from ExprImpl because we want expressions to be wrapped in an Rc,
 /// and we can't directly implement std::ops::Add, etc., for Rc&lt;E: ExprImpl&lt;Tgt;&gt;.
-#[derive(Debug,Clone)]
+#[derive(Debug, Clone)]
 pub struct Expr<T: TensorType> {
     expr: Rc<ExprImpl<T>>,
 }
 
 impl<T: TensorType> Expr<T> {
-    /// Returns the derivative of the expression with respect to the given variable.
+    pub fn new<I>(expr: I) -> Expr<T>
+    where
+        I: ExprImpl<T> + 'static,
+    {
+        Expr {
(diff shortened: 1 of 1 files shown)


src/expr.rs
```diff
@@ -51,14 +51,41 @@ pub enum OpLevel {
...
+    where
+        I: ExprImpl<T> + 'static,
+    {
+        Expr {
+            expr: Rc::new(expr),
+        }
+    }
+}
+
+impl<T: TensorType> ExprImpl<T> for Expr<T> {
```
adamcrume: I don't think Expr should implement ExprImpl, because that would allow users to do things like `Expr::new(Expr::new(something))`, which is essentially `Rc<Rc<SomeImpl>>`.  If you're trying to expose the underlying ExprImpl, the implementing Deref for Expr is probably a better approach.

adamcrume:
Out of curiosity, what other expressions are you intending to implement?

regiontog:
@adamcrume Right now I'm trying to implement MatMul.

src/expr.rs
regiontog: I've never implemented Deref before, but it seems like exactly what I need yeah.

adamcrume:
Thanks!

PR #177, opened 2018-09-13
Rewrite additon example to use Rust constructions.

Antiarchitect:
As I started to experiment with Rust bindings to TensorFlow examples was little frustrated by usage of model.pb generated by Python (!) script. An addition example is now rewritten. If this PR will be accepted I will try to modify other examples accordingly. Maybe there should be a separate example showing of how to work with model load and other stuff.


Antiarchitect:
License Agreement is signed.

adamcrume:
I know it's a bit odd, but it's set up this way on purpose.  For now, for most problems, the normal way to use this library will be to generate graphs in Python and run them in Rust.

The issue is that the TensorFlow code for computing gradients, performing gradient descent, etc. is implemented in Python.  Building any real-world ML graph without Python would require either reimplementing all of that in Rust (which would require more time than I have), or reimplementing it in C++ and binding to that in Rust (the latter half of which is what this library does).  TensorFlow is in the process of porting the gradient code from Python to C++ (see https://github.com/tensorflow/tensorflow/tree/master/tensorflow/cc/gradients), so I don't want to reimplement those things in Rust, especially since it would be temporary.

Until the upstream work is done, most users will have to rely on Python to generate non-trivial graphs.  Although simplistic, these examples are intended to showcase how the library will actually be used, so if most of our users will be relying on Python, so should the examples.

I appreciate your enthusiasm, so please don't hesitate to send other pull requests.  It's rare that I have to reject a change outright.  If you want to ask about something before writing it, you can post a message to https://groups.google.com/a/tensorflow.org/forum/#!forum/rust.

Antiarchitect:
Got it, thank you for the explanation.

Issue #149, opened 2018-03-01
Windows build: Cannot find tensorflow.lib to link


andrewcsmith:
I managed to get tensorflow 1.6.0 to build using bazel 0.11 on Windows, giving me a `.so` file that I noticed is [simply renamed to .dll](https://github.com/tensorflow/tensorflow/blob/c9dbbb898b3b8c3fcb113f5e43bc5a764ee81a8f/tensorflow/tools/ci_build/windows/libtensorflow_cpu.sh#L71****) in the CI script. I tried that, but haven't tested whether the C API works yet. 

I'm having trouble, though, because the `build.rs` script looks for a `tensorflow.lib` file rather than `tensorflow.dll`. Even when I change that one, the rust linker still tries to find `tensorflow.lib`. It doesn't work to simply rename the file; essentially, it's looking for a static lib rather than a dynamic lib.

Short question then: Can this rust binding generator recognize and link to a DLL rather than a LIB?

Happy to pay back the time in writing tutorials...been at this a little while.

adamcrume:
It'd be easy to look for a `.dll` in addition to a `.lib`.  I have little Windows experience, though, so I don't know whether that fully fixes the linking problem or not.  I'd be happy to accept a pull request if you can get it to work.  I always appreciate improvements for the Windows build.

andrewcsmith:
I seem to have fixed it. I had to create a new symbol exports table using `dumpbin /exports tensorflow.dll` and then format that properly as a `tensorflow.def` file, then run the command `lib /def:tensorflow.def /out:tensorflow.lib` to build the library. Note that this just fixes the symbols, and tensorflow will still link dynamically against the library.

I have two PRs open in [tensorflow-windows-lib](https://github.com/bekker/tensorflow-windows-lib/pulls) detailing those instructions.

This issue can be closed, but at least there's good news: tensorflow can successfully be built on windows with bazel, and linked to rust.

andrewcsmith:
I think we can't link to a `.dll` though, so we do actually need a `.lib`. I'm wondering how much of this can be automated and how much is documentation. I'll start by documenting the steps and we can make that build process automatic later.

I should also mention that this works entirely with PowerShell and the Windows `%PATH%` variable, which is (maybe) good news for anyone not working with WSL or virtual machines.

DoumanAsh:
[...]

PR #178, opened 2018-10-23
Windows: Look for existing *.dll, not *.lib

DoumanAsh:
On windows, build script looks for existing library among in PATH.
And then instructs linker to link with it using `dylib` which is for dynamic libraries.

But `*.lib` is static library, instead we should look for `*.dll` if we want to link dynamically

Related to #149

### DIFF
--- a/tensorflow-sys/build.rs
+++ b/tensorflow-sys/build.rs
@@ -67,7 +67,7 @@ fn check_windows_lib() -> bool {
 
 #[cfg(target_env = "msvc")]
 fn check_windows_lib() -> bool {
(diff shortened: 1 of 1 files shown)


DoumanAsh:
I signed it!

adamcrume:
Sorry I haven't looked at this, yet. I'm traveling, and I'll try to get to
it in the next few days.

adamcrume:
I commented on #149.

DoumanAsh:
Just tried to download C DLL, it contains only dll, no `lib` and yes, we require `*.lib` too.
So I guess PR can be closed

PR #183, opened 2019-01-15
Fix typo in addition.rs example

andrewcsmith:
A comment said `regression.py` instead of `addition.py`


andrewcsmith:
I signed it!

adamcrume:
Thanks!

PR #194, opened 2019-03-13
Update to new logo

adamcrume:
While I'm at it, are there any branding concerns that could be addressed?


ewilderj:
> While I'm at it, are there any branding concerns that could be addressed?

I'll ping you the brand usage guidelines. If there's a situation they don't address, we can ask for clarification!

PR #199, opened 2019-04-09
Update README.md

AZon8:
Its only a few targets that require large amounts of memory so its faster to build and crash then to start at --jobs=1 as far as i can tell.


adamcrume:
Closing this as obsolete.  Please send another pull request if you can sign the CLA.

PR #203, opened 2019-04-24
Add rustfmt.toml

ehsanmok:
This PR adds `rustfmt.toml` for edition 2018 and enforces the changes throughout the crate.

### DIFF
--- a/rustfmt.toml
+++ b/rustfmt.toml
@@ -1,0 +1,2 @@
+newline_style = "Unix"
+edition = "2018"
--- a/examples/addition.rs
+++ b/examples/addition.rs
@@ -1,9 +1,9 @@
 use std::error::Error;
 use std::fs::File;
 use std::io::Read;
-use std::result::Result;
 use std::path::Path;
 use std::process::exit;
+use std::result::Result;
 use tensorflow::Code;
 use tensorflow::Graph;
 use tensorflow::ImportGraphDefOptions;
@@ -13,8 +13,8 @@ use tensorflow::SessionRunArgs;
 use tensorflow::Status;
 use tensorflow::Tensor;
 
-#[cfg_attr(feature="examples_system_alloc", global_allocator)]
-#[cfg(feature="examples_system_alloc")]
+#[cfg_attr(feature = "examples_system_alloc", global_allocator)]
+#[cfg(feature = "examples_system_alloc")]
 static ALLOCATOR: std::alloc::System = std::alloc::System;
 
 fn main() {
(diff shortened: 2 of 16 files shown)


rustfmt.toml
```diff
@@ -0,0 +1,64 @@
+max_width = 100
```
adamcrume: Please remove all default values so it's clear what the non-default values are.

ehsanmok: `rustfmt.toml` was generated by `rustfmt --print-config default rustfmt.toml`. There're two other options, `current` and `minimal`. Do you mean I should go for `minimal` or keep the default and remove some of them (which ones excatly?) ?

adamcrume: I think we should start with an empty file and only set options we want to customize.  We probably want `newline_style=Unix` so we don't end up with different files having different line endings and `edition="2018"` in case someone invokes rustfmt directly without going through cargo fmt.

ehsanmok: Sounds good! I also kept major `version="One"` for consistency across.

rustfmt.toml
```diff
@@ -0,0 +1,3 @@
+newline_style = "Unix"
+edition = "2018"
+version = "One"
```
adamcrume: I'd like to specify `version`, but it's not stable.  If you remove it and reformat src/lib.rs, this PR should be good to go.

ehsanmok: Fair enough! but not sure why src/lib.rs only. For #204 don't we need to reformat all?

adamcrume: Sorry it wasn't clear.  I didn't mean to remove formatting for everything else; I meant that src/lib.rs needed to be reformatted on top of the changes already made in the pull request.  When I checked out this branch and ran `cargo fmt`, everything was clean except src/lib.rs

ehsanmok: Oh, thanks for the clarification :)

adamcrume:
I'm still seeing a diff for the definition of the impl_new macro when I run `cargo fmt`, but maybe we are running different versions of rustfmt.  (I'm on rustfmt 1.0.3-stable (d6829d6 2019-02-14)).  I'll go ahead and merge this, and if this continues to be an issue, we might need to add a `#[rustfmt::skip]`.  I think macro formatting is still being worked on.

ehsanmok:
You're right! I was using `rustfmt 1.2.1-nightly (b860fea 2019-04-18)`. I'll address it in another related PR.

PR #205, opened 2019-05-08
Adding public constructor Shape::new

functor:
`Shape` struct is public, however because its field is private there is no way to create it from outside of the module (except via `Into` etc.) This change just adds the public constructor `Shape::new` to allow construction of `Shape` from outside of its module.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1367,6 +1367,11 @@ pub fn get_registered_kernels_for_op(name: &str) -> Result<Vec<u8>> {
 pub struct Shape(Option<Vec<Option<i64>>>);
 
 impl Shape {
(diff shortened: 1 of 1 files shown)


adamcrume:
`Shape::new(data)` seems to behave exactly the same as `Shape::from(data)`.  Is there some benefit `Shape::new` provides which `Shape::from` doesn't?

By the way, I'm going on vacation, so I may not respond immediately.

functor:
Yes, `Shape::from` is equivalent, it's just a matter of public interface / expected pattern.

adamcrume:
Back from vacation. :)

I don't want to add too much redundancy to the API, but this does make it just a little less surprising.

PR #206, opened 2019-05-27
Add method to run Session with RunOptions to retrieve RunMetadata.

twuebi:
This PR adds a method `run_with_metadata` to `Session` that allows to pass a `RunOptions` protobuf to the session call. This in turn allows to retrieve `RunMetadata` which contains information on runtime and memory usage that can be passed to a `SummaryWriter` for visualization in Tensorboard. 

Note 1: the `RunOptions` proto allows to specify further options. I am not sure for what they can be used, this PR is specifically concerned with enabling the usage of Tensorboard based profiling.

Note 2: there is only a single test that checks that the session can be run with serialized `RunOptions` with trace level set to `RunOptions_TraceLevel::FULL_TRACE`. The file `test_resources/session/run_options/run_options.bin` contains said protobuf. I did manually verify that the `&[u8]` returned by `run_with_metadata` is a valid RunMetadata proto by decoding it using this [crate](https://github.com/danieldk/sticker/tree/master/tf-proto) and [protobuf](https://crates.io/crates/protobuf).

```rust
...
let meta_buf = session.run_with_metadata(&mut step, &run_options).unwrap();
let mut run_metadata = RunMetadata::new();
run_metadata.merge_from(&mut protobuf::CodedInputStream::from_bytes(&meta_buf));
```

It produces the expected [output](https://github.com/tensorflow/rust/files/3224419/output.txt).

____________________________________________________________________________

For usage with Tensorboard, one needs to use the `tf.contrib.summary` api that moves the summary writing into the graph. The `RunMetadata` proto needs to wrapped in a `TaggedRunMetadata` proto which again has to wrapped in an `Event` proto. One then needs to run the session with the target_op defined by `tensorflow.python.ops.summary_ops_v2.import_event` while feeding the serialized `Event` proto to a string placeholder.

### DIFF
--- a/src/session.rs
+++ b/src/session.rs
@@ -154,14 +154,30 @@ impl Session {
     pub fn run(&self, step: &mut SessionRunArgs<'_>) -> Result<()> {
         // In case we're running it a second time and not all outputs were taken out.
         step.drop_output_tensors();
+        // make sure run_metadata is either None or an empty TF_Buffer
+        step.maybe_reset_run_metadata();
 
         let mut status = Status::new();
         let maybe_tensors: Result<_> = step.input_tensors.iter().map(|t| t.inner()).collect();
         let input_tensors: Vec<_> = maybe_tensors?;
+        let run_options_ptr = match step.run_options.as_ref() {
+            Some(buf) => buf.inner(),
+            None => ptr::null(),
+        };
+
+        let mut run_metadata_buf = if step.request_metadata {
+            Some(unsafe { Buffer::new_unallocated() })
+        } else {
+            None
+        };
+        let run_metadata_ptr = match run_metadata_buf.as_mut() {
+            Some(meta) => meta.inner_mut(),
+            None => ptr::null_mut(),
+        };
         unsafe {
             tf::TF_SessionRun(
                 self.inner,
-                ptr::null(),
+                run_options_ptr,
                 step.input_ports.as_ptr(),
                 input_tensors.as_ptr() as *const *const tf::TF_Tensor,
                 input_tensors.len() as c_int,
@@ -170,10 +186,12 @@ impl Session {
                 step.output_tensors.len() as c_int,
                 step.target_operations.as_mut_ptr(),
                 step.target_operations.len() as c_int,
-                ptr::null_mut(),
+                run_metadata_ptr,
                 status.inner(),
             );
-        };
+            step.run_metadata = run_metadata_buf.map(Into::into);
+        }
+
         status.into_result()
     }
 
@@ -274,6 +292,10 @@ pub struct SessionRunArgs<'l> {
 
     target_operations: Vec<*const tf::TF_Operation>,
 
+    run_options: Option<Buffer<u8>>,
+    run_metadata: Option<Vec<u8>>,
+    request_metadata: bool,
+
     phantom: marker::PhantomData<&'l ()>,
 }
 
@@ -286,6 +308,10 @@ impl<'l> SessionRunArgs<'l> {
 
             output_ports: vec![],
             output_tensors: vec![],
+
+            run_options: None,
+            run_metadata: None,
+            request_metadata: false,
 
             target_operations: vec![],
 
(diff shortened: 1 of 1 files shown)


src/session.rs
```diff
@@ -521,6 +565,24 @@ mod tests {
         assert_eq!(output_tensor[1], 6.0);
     }
 
+    #[test]
+    fn test_run_metadata() {
+        let (session, x_operation, y_operation) = create_session();
+        let mut x = <Tensor<f32>>::new(&[2]);
+        x[0] = 2.0;
+        x[1] = 3.0;
```
adamcrume: I know we don't have examples, but you can replace the previous three lines with the shorter and less error-prone (although slightly odd):

```
let x = Tensor::<f32>::from(&[2.0, 3.0][..]);
```

(The weird `[..]` is because we have a `From` impl for slices but not arrays, and for some reason the compiler doesn't catch on that `&[2.0, 3.0]` can be treated as a slice as well as a reference to an array.  This is a shorthand for treating it as a slice.)

src/session.rs
```diff
@@ -521,6 +565,24 @@ mod tests {
...
+        let mut step = SessionRunArgs::new();
+        step.add_feed(&x_operation, 0, &x);
+        let output_token = step.request_fetch(&y_operation, 0);
+        // read serialized RunOptions protobuf with trace level set to full.
+        let f = fs::read("test_resources/session/run_options/run_options.bin").unwrap();
```
adamcrume: I don't think a separate file is necessary for this, especially since it looks like an empty protobuf anyway.  You could just hard-code the array here.

src/session.rs
```diff
@@ -177,6 +177,49 @@ impl Session {
...
+        // In case we're running it a second time and not all outputs were taken out.
+        step.drop_output_tensors();
+
+        let mut status = Status::new();
+        let maybe_tensors: Result<_> = step.input_tensors.iter().map(|t| t.inner()).collect();
+        let input_tensors: Vec<_> = maybe_tensors?;
+        let run_metadata = unsafe {
+            let mut run_metadata: Buffer<u8> = Buffer::new_unallocated();
+            let run_options =
+                tf::TF_NewBufferFromString(run_options.as_ptr() as *const _, run_options.len());
```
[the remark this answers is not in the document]
twuebi: Thanks! Replaced with `Buffer::from_ptr`.

src/session.rs
```diff
@@ -177,6 +177,49 @@ impl Session {
...
 
+    /// Runs the graph with RunMetadata, feeding the inputs and then fetching the outputs
+    /// requested in the step.  Note that the session has interior mutability;
+    /// this may mutate variables in the graph, and the caller is responsible
+    /// for handling race conditions.
+    /// `run_options` should be a serialized [`RunOptions` proto](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/protobuf/config.proto).
+    /// The returned [`RunMetadata` proto](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/protobuf/config.proto)
+    /// is a serialized proto that holds information about runtime and memory usage depending on the
+    /// trace level specified in `run_options`.
+    pub fn run_with_metadata(
```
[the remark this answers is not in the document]
twuebi: I removed the method and moved the two buffers as options into `SessionRunArgs`. The metadata can be requested via `SessionRunArgs::set_request_metadata(bool)`. Setting it to true constructs the unallocated metadata Buffer that is then used in `Session::run`.

src/session.rs
twuebi: Thanks! Replaced with a slice.

src/session.rs
twuebi: Proto is now hardcoded.

src/session.rs
```diff
@@ -422,6 +479,14 @@ impl<'l> SessionRunArgs<'l> {
             *tensor = ptr::null_mut();
         }
     }
+
+    fn maybe_reset_run_metadata(&mut self) {
+        if let Some(t) = self.run_metadata.as_ref() {
+            if !t.inner().is_null() {
+                self.run_metadata = unsafe { Some(Buffer::new_unallocated()) };
```
adamcrume: This isn't safe, because unallocated buffers are in an unsafe state and shouldn't exist outside of unsafe blocks.  A safer way to do this is to create it inside `Session::run`'s unsafe block and assign it to `SessionRunArgs::run_metadata` after it's been filled.

src/session.rs
```diff
@@ -411,6 +429,45 @@ impl<'l> SessionRunArgs<'l> {
         }
     }
 
+    /// Sets the `RunOptions`. `run_options` is a serialized [`RunOptions` proto](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/protobuf/config.proto).
+    pub fn set_run_options(&mut self, run_options: &[u8]) {
+        self.run_options = unsafe {
+            Some(Buffer::from_ptr(
```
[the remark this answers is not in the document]
twuebi: Replaced with `Buffer::from`.

src/session.rs
```diff
@@ -154,14 +154,26 @@ impl Session {
...
         unsafe {
+            let run_options_ptr = match step.run_options.as_ref() {
+                Some(buf) => buf.inner(),
+                None => ptr::null(),
+            };
+            let run_metadata = match step.run_metadata.as_mut() {
+                Some(meta) => meta.inner_mut(),
+                None => ptr::null_mut(),
+            };
+
```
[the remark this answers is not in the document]
twuebi: Moved them outside.

src/session.rs
twuebi: The metadata buffer is now created in `Session::run` and converted to `Vec<u8>` before being stored in `SessionRunArgs::run_metadata`.

With the changes above and the run_metadata buffer becoming a local variable, we either need to create the pointer inside the unsafe block or need a second unsafe block exclusively for the creation of the metadata buffer. I've gone for the latter as it keeps the pointer creation outside the unsafe block. If there's something speaking against it, I'll be happy to go for the other option.

src/session.rs
```diff
@@ -287,6 +309,10 @@ impl<'l> SessionRunArgs<'l> {
             output_ports: vec![],
             output_tensors: vec![],
 
+            run_options: None,
+            run_metadata: None,
+            return_metadata: true,
```
twuebi: This one sneaked in, changing the default to false.

src/session.rs
```diff
@@ -411,6 +437,34 @@ impl<'l> SessionRunArgs<'l> {
...
+    }
+
+    /// Requests `run_metadata`. The serialized [`RunMetadata` proto](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/protobuf/config.proto)
+    /// can be retrieved via `self::get_metadata` after calling `Session::run`.
+    pub fn set_request_metadata(&mut self, request: bool) {
+        self.return_metadata = request;
+    }
+
+    /// Returns whether `RunMetadata` should be stored.
+    pub fn is_run_metadata_requested(&self) -> bool {
```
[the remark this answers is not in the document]
twuebi: Changed name to `is_request_metadata`, also renamed the struct member `return_metadata` to `request_metadata`.

adamcrume:
Thanks!

PR #212, opened 2019-06-24
Create basic ops module

adamcrume:
This creates an `ops` module with a few basic operations implemented.  This is the first pull request in a sequence intended to support defining and training models entirely in Rust (no Python necessary).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -25,6 +25,8 @@ random = "0.12.2"
 [features]
 tensorflow_gpu = ["tensorflow-sys/tensorflow_gpu"]
 tensorflow_unstable = []
+# Enables the new ops module which supports building graphs with less boilerplate.
+experimental_training = []
 # This is for testing purposes; users should not use this.
 examples_system_alloc = ["tensorflow-sys/examples_system_alloc"]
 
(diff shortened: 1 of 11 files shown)


adamcrume:
Thanks!

I was hoping to autogenerate the ops eventually.  I wanted to get enough of the Rust API implemented to enable writing a simple end-to-end example first, and then improve the implementation.  It may force me to pick a Rust proto library, which I avoided in the past since there was no clear best library.

I haven't had time to look at the eager API, until now.  In tensorflow/c/eager/c_api.h, I see:

```
// WARNING: Unlike tensorflow/c/c_api.h, the API here is not guaranteed to be
// stable and can change without notice.
```

Is that temporary, or will it always be unstable?  Is the old API planned to be deprecated/removed/unsupported, or is new functionality just going to be added to the eager API?  I'm also concerned with this comment:

```
// TODO(ashankar): Merge with TF_Session?
typedef struct TFE_Context TFE_Context;
```

That would be a big change, and I'm not sure I could hide that change in the Rust API, so I'd prefer that to be answered before implementing the eager API in Rust.

adamcrume:
I'll go ahead and merge this, but I'd like to follow up later about the eager API.

PR #217, opened 2019-07-23
set rustc-link-lib/search for tensorflow lib found via pkg_config

afloren:
Currently, if a Tensorflow library is installed in a non-system folder (e.g., /opt/tensorflow/lib) but still discoverable via pkg_config then dependent crates will fail to link at compile time.

This sets rustc-link-lib and rustc-link-search values for the library discovered by pkg_config.


adamcrume:
Great, thanks!

PR #218, opened 2019-07-30
Revert "set rustc-link-lib/search for tensorflow lib found via pkg_config"

afloren:
Reverts tensorflow/rust#217

On closer inspection, `find_library` does print these by default. I was hitting an unrelated bug with pkg-config itself. Apologies for the confusion.

### DIFF
--- a/tensorflow-sys/build.rs
+++ b/tensorflow-sys/build.rs
@@ -43,13 +43,7 @@ fn main() {
         return;
     }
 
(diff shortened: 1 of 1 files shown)


adamcrume:
Ew, can't say I like pkg_config's default behavior, there.  Would you mind adding a comment explaining what pkg_config does and replacing `find_library` (which appears to be deprecated) with `probe_library`?

PR #223, opened 2019-10-14
Added set and get to the Tensor

### LINKED ISSUE #222: Set data to a Tensor in Row/Column style
I can create a 2x2 Matrix in this way:
```
let mut matrix = Tensor::new(&[2, 2]);
```
I would like to set its data in row / column style, something like:
```
// Using a macro to generate the Index
matrix[get_index(1, 1)] = 1.0;

// Using a function
matrix.set(1, 1, 1.0);
```

I'm not able to find any function / macro to perform this kind of operation, there is a reason for this?
(In case I'm available to submit a PR to implement it).

AndreaCatania:
Added set and get to the Tensor, closes #222.

You should consider to use usize instead of u64 for the Tensor dims, so we don't need to convert u64 to usize.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1211,6 +1211,60 @@ impl<T: TensorType> Tensor<T> {
         Ok(self)
     }
 
+    /// Set one single value on the tensor.
+    ///
+    /// ```
+    /// # use tensorflow::Tensor;
+    /// let mut a = Tensor::<i32>::new(&[3, 3, 3]);
+    ///
+    /// a.set(&[0, 0, 1], 10);
+    /// assert_eq!(a[0 + 0 + 1], 10);
+    ///
+    /// a.set(&[2, 2, 0], 9);
+    /// assert_eq!(a[2*9 + 2*3 + 0], 9);
+    /// ```
+    pub fn set(&mut self, indices: &[u64], value: T) {
+        let index = self.get_index(indices);
+        self[index] = value;
+    }
+
+    /// Get one single value from the Tensor.
+    ///
+    /// ```
+    /// # use tensorflow::Tensor;
+    /// let mut a = Tensor::<i32>::new(&[2, 3, 5]);
+    ///
+    /// a[1*15 + 1*5 + 1] = 5;
+    /// assert_eq!(a.get(&[1, 1, 1]), 5);
+    /// ```
+    pub fn get(&self, indices: &[u64]) -> T {
+        let index = self.get_index(indices);
+        self[index].clone()
+    }
+
+    /// Get the array index from rows / columns indices.
+    ///
+    /// ```
+    /// # use tensorflow::Tensor;
+    /// let a = Tensor::<f32>::new(&[3, 3, 3]);
+    ///
+    /// assert_eq!(a.get_index(&[2, 2, 2]), 26);
+    /// assert_eq!(a.get_index(&[1, 2, 2]), 17);
+    /// assert_eq!(a.get_index(&[1, 2, 0]), 15);
+    /// assert_eq!(a.get_index(&[1, 0, 1]), 10);
+    /// ```
+    pub fn get_index(&self, indices: &[u64]) -> usize {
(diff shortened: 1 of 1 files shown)


AndreaCatania:
@googlebot I signed it!

src/lib.rs
```diff
@@ -1211,6 +1211,60 @@ impl<T: TensorType> Tensor<T> {
...
+    /// # use tensorflow::Tensor;
+    /// let a = Tensor::<f32>::new(&[3, 3, 3]);
+    ///
+    /// assert_eq!(a.get_index(&[2, 2, 2]), 26);
+    /// assert_eq!(a.get_index(&[1, 2, 2]), 25);
+    /// assert_eq!(a.get_index(&[1, 2, 0]), 7);
+    /// assert_eq!(a.get_index(&[1, 0]), 1);
+    /// ```
+    pub fn get_index(&self, indices: &[u64]) -> usize {
+        assert!(self.dims.len() >= indices.len());
```
adamcrume: I'm not sure there's any real benefit to allowing users to specify only part of the index, and it's likely to cause confusion.  I would just require `self.dims.len() == indices.len()`.

src/lib.rs
```diff
@@ -1211,6 +1211,60 @@ impl<T: TensorType> Tensor<T> {
...
+        let mut index = 0;
+        let mut d = 1;
+        for i in 0..indices.len() {
+            assert!(self.dims[i] > indices[i]);
+            index += indices[i] * d;
```
adamcrume: The docs on `Tensor` say that element 1 corresponds to index [0, ..., 1], but this code is mapping element 1 to index [1, ..., 0].

adamcrume:
I'm using u64 for the dimensions because the underlying TensorFlow C API uses int64_t for dimensions.  I use unsigned instead of signed whenever we know that -1 isn't supported, but at least we don't have to worry about int64_t versus size_t issues with the C API.  It does make the Rust API a little funny, but I'm trying to avoid edge cases where the C API works but the Rust API doesn't.

src/lib.rs
AndreaCatania: Yep I agree, done

AndreaCatania:
> This could use some tests. For example, creating a [2, 3, 5]-shape tensor, setting element [1, 1, 1] with `set`, and verifying that the element at index 1+1_5+1_5*3 =21 is set.

I've modified the previous test with the above one, you can find it in the `get` function.

> I'm using u64 for the dimensions because the underlying TensorFlow C API uses int64_t for dimensions.

The things is that you can't access any array (an so a tensor) using a type different than a `usize`, this mean that at some point you have to always convert u64 back to usize.

With the actual implementation you can create Tensor of u64 size even in a 16 or 32 bit architectures (where the usize is 16 or 32 bits) this mean that part of the indices will be truncated.

On the other hand, converting a usize to u64 is always possible and cause less problem (since you can't anymore create Tensor of unsupported sizes). So I think that the conversion should be handled safely internally by this crate.

adamcrume:
Hm, you're right that users can't actually fit a full, materialized 64-bit-sized tensor in memory on a 32-bit architecture.  I was concerned about the indexes used here being the same type as what's returned by Tensor::dims, but users are already using usize to index into tensors right now, which is the more natural type.  Feel free to change it to usize.

AndreaCatania:
I think that for now it's better to use `u64` for setters and getters too, but I think that in future the Tensor should use only `usize` and perform the conversion to `u64` internally.

adamcrume:
We can't change the Tensor dimensions from u64 to usize without making a breaking change, unless we simply add separate methods.  It would also mean a mismatch between Tensor dimension types and dimension types for tensors in the graph which are not fully materialized on the current machine.  We also have to consider sparse tensors, which could easily have a u64 address space, even if only a couple of items are filled out.

This is something we'll need to revisit before moving to version 1.0.

adamcrume:
Thanks for the contribution!

AndreaCatania:
Welcome and Thank you!

PR #230, opened 2019-11-22
Use `std::alloc` instead of `aligned_alloc`

jonas-schievink:
`aligned_alloc` is my most downloaded crate, only because the Tensorflow bindings are using it. They should not, since the crate is slow, wasteful, and buggy. Fortunately, `std::alloc` now has enough stable API surface to replace `aligned_alloc` here.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -12,7 +12,6 @@ edition = "2018"
 
 [dependencies]
 libc = "0.2.62"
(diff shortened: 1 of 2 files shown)


jonas-schievink:
@googlebot I signed it!

adamcrume:
Thanks!  I appreciate the cleanup.

PR #240, opened 2020-02-18
RecordReader for TFRecords

masonk:
I guess I left the work unfinished when I submitted RecordWriter a few years ago. I was hoping some upstream issues would be resolved (e.g., GATs for efficient iteration, and I was expecting the Read trait to change to deal with the uninitialized buffers problem).

I think this is the best that can be done for right now, and it's better to have something than nothing.

### DIFF
--- a/src/io.rs
+++ b/src/io.rs
@@ -154,11 +154,8 @@ where
     /// Use this to find out how large the byte slice needs to be to read the next record.
     pub fn peek_next_len(&mut self) -> Result<Option<u64>, RecordReadError> {
         let len = self.read_next_len_unchecked()?;
-        match len {
-            Some(_len) => {
-                self.reader.seek(SeekFrom::Current(-8))?;
-            }
-            _ => {}
+        if let Some(_) = len {
+            self.reader.seek(SeekFrom::Current(-8))?;
         }
 
         Ok(len)
@@ -237,21 +234,19 @@ where
             self.reader.seek(SeekFrom::Current(8 + len as i64))?;
             return Err(RecordReadError::BufferTooShort { needed: len });
         }
+
         let mut len_bytes = [0u8; 8];
         LittleEndian::write_u64(&mut len_bytes, len);
-        let len_ok = self.checksum(&mut len_bytes)?;
+        if !self.checksum(&mut len_bytes)? {
+            return Err(RecordReadError::CorruptFile);
+        }
 
         let slice = &mut buf[0..len as usize];
         self.read_bytes_exact_unchecked(slice)?;
-        let bytes_ok = self.checksum(slice)?;
 
-        if bytes_ok && len_ok {
-            return Ok(Some(len as usize));
-        } else {
-            if !len_ok {
-                return Err(RecordReadError::CorruptFile);
-            }
-            return Err(RecordReadError::CorruptRecord);
+        match self.checksum(slice)? {
+            true => Ok(Some(len as usize)),
+            false => Err(RecordReadError::CorruptRecord),
         }
     }
     /// Allocate a Vec<u8> on the heap and read the next record into it.
@@ -293,19 +288,16 @@ where
             Some(len) => len,
             None => return Ok(None),
         };
-        let mut vec = vec![0u8; len as usize];
         let mut len_bytes = [0u8; 8];
         LittleEndian::write_u64(&mut len_bytes, len);
-        let len_ok = self.checksum(&mut len_bytes)?;
+        if !self.checksum(&mut len_bytes)? {
+            return Err(RecordReadError::CorruptFile);
+        }
+        let mut vec = vec![0u8; len as usize];
         self.read_bytes_exact_unchecked(&mut vec)?;
-        let bytes_ok = self.checksum(&vec)?;
-        if bytes_ok && len_ok {
-            return Ok(Some(vec));
-        } else {
-            if !len_ok {
-                return Err(RecordReadError::CorruptFile);
-            }
-            return Err(RecordReadError::CorruptRecord);
+        match self.checksum(&vec)? {
+            true => Ok(Some(vec)),
+            false => Err(RecordReadError::CorruptRecord),
         }
     }
     /// Convert the Reader into an Iterator<Item = Result<Vec<u8>, RecordReadError>, which iterates
@@ -430,10 +422,9 @@ mod tests {
                     Some(len) => assert_eq!(&ary[0..len], records[i].as_bytes()),
                     None => break,
                 },
-                Err(RecordReadError::CorruptFile) | Err(RecordReadError::IoError { .. }) => {
-                    break;
+                Err(e @ _) => {
+                    panic!("Received an unexpected error: {:?}", e);
                 }
-                _ => {}
             }
             i += 1;
         }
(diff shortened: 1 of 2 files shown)


masonk:
BTW I'm a Googler. Do I still need to sign this with my new email address?

src/io.rs
```diff
@@ -32,30 +40,147 @@ where
...
+    }
+    /// Allocate a Vec<u8> on the heap and read the next record into it.
+    /// Returns the filled Vec, if successful.
+    /// Returns None, if it could read exactly 0 bytes (indicating EOF)
+    pub fn read_next_owned(&mut self) -> Result<Option<Vec<u8>>, RecordReadError> {
+        let len = match self.read_next_len()? {
+            Some(len) => len,
+            None => return Ok(None),
+        };
+        let mut vec = vec![0u8; len as usize];
```
masonk: I didn't bother with reserving capacity and unsafing to write into uninitialized memory, because I think that performance conscious users are going to use the other method anyway. But let me know what you think.

src/io.rs
```diff
@@ -32,30 +40,147 @@ where
...
+        self.read_bytes_exact(len, &mut vec)?;
+
+        Ok(Some(vec))
+    }
+    /// Convert the Reader into an Iterator<Item = Result<Vec<u8>, RecordReadError>, which iterates
+    /// the whole file.
+    pub fn into_iter_owned(self) -> impl Iterator<Item = Result<Vec<u8>, RecordReadError>> {
```
masonk: To my knowledge, without GATs, there is no way to write an iterator that owns an internal buffer for reading. Let me know if I'm missing something.

masonk:
I registered my github account with Google using the corp self-service tool, but that didn't make this bot happy.

src/io.rs
```diff
@@ -84,14 +209,41 @@ mod tests {
...
+            assert_eq!(actual, expected);
+        }
 
-        let mut af = File::open(actual_filename).unwrap();
-        let mut ef = File::open(expected_filename).unwrap();
-        let mut actual = vec![0; 0];
-        let mut expected = vec![0; 0];
-        af.read_to_end(&mut actual).unwrap();
-        ef.read_to_end(&mut expected).unwrap();
+        let _ = std::fs::remove_file(actual_filename);
```
masonk: sneak in a line to clean up test files

liufuyang: I am not sure but perhaps we can also create some file in some temp folder? But this seems working as well.

src/io.rs
liufuyang: Hmm, I think my knowledge on Rust and TF is not enough yet to give an good answer. Do you perhaps have some other people you can ask for a good review? (Or perhaps we should ask in the community?)

Also I don't have much time recently. Otherwise I will try give a closer look.

masonk:
I squashed these commits, added usage examples in the comments, and changed the email to something that makes cla happy. Please take another look at this PR.

Oh, and I consolidated `InvalidLengthChecksum` and `InvalidContentChecksum` into `InvalidChecksum`, as I couldn't imagine a scenario in which a user would do something different between those cases.

src/io.rs
```diff
@@ -32,30 +37,181 @@ where
...
+        };
+
+        let mut len_bytes = [0u8; 8];
+        LittleEndian::write_u64(&mut len_bytes, len);
+
+        let expected_len_crc32 = self.reader.read_u32::<LittleEndian>()?;
+        let actual_len_crc32 = mask(crc32::checksum_castagnoli(&len_bytes));
+        if expected_len_crc32 != actual_len_crc32 {
+            return Err(RecordReadError::InvalidChecksum);
+        }
```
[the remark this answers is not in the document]
masonk: Done.

src/io.rs
```diff
@@ -32,30 +37,181 @@ where
...
+    /// loop {
+    ///     let next = records.read_next(&mut buf);
+    ///     match next {
+    ///         Ok(res) => match res {
+    ///             Some(len) => train(&buf[0..len]),
+    ///             None => break,
+    ///         }
+    ///         Err(e) => { warn!("{:?}", e); break }
+    ///     }
+    /// }
```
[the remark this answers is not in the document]
masonk: I converted my existing test into a doctest and converted these examples into doctests.

src/io.rs
```diff
@@ -32,30 +37,181 @@ where
...
+    pub fn read_next(&mut self, buf: &mut [u8]) -> Result<Option<u64>, RecordReadError> {
```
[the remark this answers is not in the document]
masonk: Good catch.

I went with the following approach:

+ pub fn `peek_next_len()`, which tells the length of the next record, and can be used to resize a heap buffer in advance of the `read_next` call.  Rather than wrap into a `BufRead` internally, I added `Seek` bounds, giving control to the caller. In most cases they will be passing `BufRead`s anyway.
+ `read_next()` skips records which are longer than the supplied buffer. So if you don't know required buffer size, peek first

src/io.rs
```diff
@@ -32,30 +39,323 @@ where
...
+    ///     }
+    ///     i += 1;
+    /// }
+    /// ```
+    pub fn read_next_owned(&mut self) -> Result<Option<Vec<u8>>, RecordReadError> {
+        let len = match self.read_next_len_unchecked()? {
+            Some(len) => len,
+            None => return Ok(None),
+        };
+        let mut vec = vec![0u8; len as usize];
```
adamcrume: The buffer shouldn't be allocated until after the length is checksummed, or the program could easily try to allocate a terabyte sized buffer unnecessarily.

src/io.rs
```diff
@@ -32,30 +39,323 @@ where
...
+        let bytes_ok = self.checksum(slice)?;
+
+        if bytes_ok && len_ok {
+            return Ok(Some(len as usize));
+        } else {
+            if !len_ok {
+                return Err(RecordReadError::CorruptFile);
+            }
+            return Err(RecordReadError::CorruptRecord);
+        }
```
adamcrume: This could be simplified to:

```
if !len_ok {
  Err(RecordReadError::CorruptFile)
} else if !bytes_ok {
  Err(RecordReadError::CorruptRecord)
} else {
  Ok(Some(len as usize))
}
```

src/io.rs
```diff
@@ -84,14 +384,95 @@ mod tests {
...
+            }
+
+            let next = reader.read_next(&mut ary);
+            match next {
+                Ok(res) => match res {
+                    Some(len) => assert_eq!(&ary[0..len], records[i].as_bytes()),
+                    None => break,
+                },
+                Err(RecordReadError::CorruptFile) | Err(RecordReadError::IoError { .. }) => {
+                    break;
```
adamcrume: Shouldn't this fail the test?

src/io.rs
```diff
@@ -84,14 +384,95 @@ mod tests {
...
+                    Err(RecordReadError::BufferTooShort { needed: v }) => v == 4,
+                    _ => false,
+                };
+                assert!(buffer_too_short);
+            } else if i == 2 {
+                assert_eq!(next.unwrap().unwrap(), 3);
+                assert_eq!(&ary, records[i].as_bytes());
+            } else {
+                break;
+            }
```
adamcrume: Rather than a loop with an if-else that does something different for each index, it would make more sense to unroll it.

src/io.rs
```diff
@@ -32,30 +39,323 @@ where
...
+        self.read_bytes_exact_unchecked(&mut vec)?;
+        let bytes_ok = self.checksum(&vec)?;
+        if bytes_ok && len_ok {
+            return Ok(Some(vec));
+        } else {
+            if !len_ok {
+                return Err(RecordReadError::CorruptFile);
+            }
+            return Err(RecordReadError::CorruptRecord);
+        }
```
adamcrume: This could be simplified to:

```
if !len_ok {
  Err(RecordReadError::CorruptFile)
} else if !bytes_ok {
  Err(RecordReadError::CorruptRecord)
} else {
  Ok(Some(len as usize))
}
```

src/io.rs
masonk: Heh. I think I was a little tired last night.

adamcrume:
Looks good, thanks a lot!

masonk:
Thanks for the great reviews

liufuyang:
Thank you both very much on this 😄

PR #247, opened 2020-03-31
Update downloaded libraries to matching version (1.15)

Darktrop:
During the upgrade to 1.14 and 1.15 the libraries where not upgraded in tensorflow-sys crate

### DIFF
--- a/tensorflow-sys/build.rs
+++ b/tensorflow-sys/build.rs
@@ -4,8 +4,9 @@ extern crate pkg_config;
 extern crate semver;
 extern crate tar;
 
+use std::env::consts::DLL_EXTENSION;
 use std::error::Error;
-use std::fs::File;
+use std::fs::{DirEntry, File};
 use std::io::BufWriter;
 use std::io::Write;
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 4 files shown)


Darktrop:
@googlebot I signed it!

kahlil29:
Does this mean that once this PR is merged, the `tensorflow` crate will automatically pull in version 1.15 of tensorflow if it does not detect a pre-existing system installation?

Darktrop:
That's right. When tensorflow-sys crate build, it checks if there is pre-existing installation using pkgconfig. If not it will download version 1.15 of tensorflow. Currently (on master It is fetching version 1.13.1.

kahlil29:
> Currently (on master) It is fetching version 1.13.1. 
Yes, and that's leading to a fair bit of problems for me. 
Thanks for this PR, kind sir @Darktrop 
Cannot wait for it to be merged and released.

tensorflow-sys/build.rs
```diff
@@ -164,43 +166,34 @@ fn install_prebuilt() {
...
-            "File {} already exists, deleting.",
-            new_framework_library_full_path.display()
-        );
-        std::fs::remove_file(&new_framework_library_full_path).unwrap();
-    }
-    let new_library_full_path = output.join(&library_file);
-    if new_library_full_path.exists() {
+
+    let framework_files =
+        glob::glob(&format!("{}/lib*.{}", lib_dir.display(), DLL_EXTENSION)).unwrap();
```
adamcrume, line 171: This misses libtensorflow.so.1.15.0 and libtensorflow.so.1, and libtensorflow.so.1 is what's being linked against.  Likewise for libtensorflow_framework.

Darktrop: Indeed the code above, It will take libtensorflow.1.15.0.so but not libtensorflow.so.1.15.0. I did not expect to have the version after the extension.

Darktrop:
I've run the script `./test-all` on macos (version 10.14) and centos 7
It is passing on both.

tensorflow-sys/build.rs
Darktrop: I have removed glob dependency and list all file in lib_dir.

tensorflow-sys/build.rs
```diff
@@ -164,43 +166,34 @@ fn install_prebuilt() {
...
+    for path in framework_files.filter_map(Result::ok) {
+        let library_full_path: PathBuf = path;
+        let new_library_full_path =
+            output.join(&PathBuf::from(library_full_path.file_name().unwrap()));
```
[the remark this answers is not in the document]
adamcrume: Can you remove the extra PathBuf?

adamcrume:
Thanks a lot!

PR #250, opened 2020-04-20
Allow tensorflow-sys to download prebuilt windows releases.

photex:
Howdy!

This appears to work as intended, but please let me know what you'd like to have changed. Since there are pre-built releases available for windows I figured it'd be worthwhile to also support them in this crate.

Cheers!
Chip

### DIFF
--- a/tensorflow-sys/Cargo.toml
+++ b/tensorflow-sys/Cargo.toml
@@ -23,6 +23,7 @@ flate2 = "1.0.12"
 pkg-config = "0.3.16"
 semver = "0.9.0"
 tar = "0.4.26"
+zip = "0.5"
 
 [features]
 tensorflow_gpu = []
--- a/tensorflow-sys/build.rs
+++ b/tensorflow-sys/build.rs
@@ -4,20 +4,24 @@ extern crate pkg_config;
 extern crate semver;
 extern crate tar;
 
-use std::env::consts::DLL_EXTENSION;
+use std::env::{
+    self,
+    consts::{DLL_EXTENSION, DLL_PREFIX},
+};
 use std::error::Error;
-use std::fs::File;
-use std::io::BufWriter;
-use std::io::Write;
+use std::fs::{self, File};
+use std::io::{self, BufWriter, Write};
 use std::path::{Path, PathBuf};
-use std::process;
-use std::process::Command;
-use std::{env, fs};
+use std::process::{self, Command};
 
 use curl::easy::Easy;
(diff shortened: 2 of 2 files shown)


tensorflow-sys/build.rs
```diff
@@ -94,13 +97,43 @@ fn remove_suffix(value: &mut String, suffix: &str) {
...
+fn extract<P: AsRef<Path>, P2: AsRef<Path>>(archive_path: P, extract_to: P2) {
+    fs::create_dir_all(&extract_to)
+        .expect("Failed to create output path for zip archive.");
+    let file = File::open(archive_path)
+        .expect("Unable to open libtensorflow zip archive.");
+    let mut archive = ZipArchive::new(file).unwrap();
+    for i in 0..archive.len() {
+        let mut zipfile = archive.by_index(i).unwrap();
+        let output_path = extract_to.as_ref().join(zipfile.sanitized_name());
+        if (&*zipfile.name()).starts_with("lib") {
```
adamcrume, line 118: Can't this be simplified to just `zipfile.name().starts_with("lib")`?

adamcrume:
Please ensure that test-all runs successfully.  (There's supposed to be a Travis CI check, but it's not running for some reason.)

tensorflow-sys/Cargo.toml
```diff
@@ -20,6 +20,7 @@ libc = "0.2.43"
 [build-dependencies]
 curl = "0.4.24"
 flate2 = "1.0.12"
+zip = "0.5"
```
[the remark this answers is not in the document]
photex: No problem

tensorflow-sys/build.rs
photex: It sure can. Can't recall what sequence of silliness led me to that anyway. :D

photex:
Since I'm attempting to build and work with this without mingw or bash etc I'll have to do the work of ./test-all by hand.
I'm happy to put together a powershell or batch file if you think that's more suitable.

I also just realized that when I tested my changes I didn't actually try to use the crate in my app (as in `use tensorflow`, so this current state is incorrect. 

Changes incoming.

photex:
Alrighty @adamcrume I ran the work in ./test-all manually in Windows and everything worked.

I then ran it in WSL and things worked fine as well. I don't have a MacOS machine to test with otherwise I'd be happy to also check there.

adamcrume:
Thanks!  I always appreciate maintenance for the Windows build.

PR #259, opened 2020-06-13
OutputName::from_str - Allow 0 colons in name, default to index 0

kykosic:
Currently, the `OutputName::from_str` method requires exactly 1 colon in the name to be valid, with syntax "name:index". I'm proposing that if just "name" is given with no index, then the index should default to 0.

I'm proposing this change as I recently exported a simple tf.keras sequential model from python, and when I attempt to load it with `SaveModelBundle::load`, it errors. The names of the outputs in the saved model are:
```
NoOp
serving_default_inputs:0
StatefulPartitionedCall:0
```
The model fails to load due to "NoOp" not having an index. This pull request fixes this issue, and the model loads+runs correctly.

If there is an alternative suggestion on how to handle this, I'd be happy to implement that as well. We could also make the index an Optional<c_int> instead of c_int.

### DIFF
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -1760,13 +1760,14 @@ impl FromStr for OutputName {
     type Err = Status;
     fn from_str(s: &str) -> Result<Self> {
         let splits: Vec<_> = s.split(':').collect();
-        if splits.len() != 2 {
-            return Err(Status::new_set_lossy(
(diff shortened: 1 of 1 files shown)


adamcrume:
I had a very similar local change that somehow didn't get into the repo.  Thanks!

Issue #166, opened 2018-06-25
Wrap TF_GetOpList


adamcrume:
We have a TODO to wrap TF_GetOpList: https://github.com/tensorflow/rust/blob/a5409d183b2a7301ca946a5d2d13d34e471244ef/src/lib.rs#L1240

Doing this properly requires figuring out proto support, first.

kykosic:
I'd like to take this issue if available. I've already started working on it, but I'm not 100% clear on what to do once I have an `OpList` proto. Are we just storing it as a member of the `Library` struct, matching the pattern in [the C library](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/core/framework/load_library.cc)?

adamcrume:
Yeah, storing it as a member of the `Library` struct should work.  I'm currently trying to keep our choice of proto library as an implementation detail, so the types used in the public API need to be hand-written protos.

allentv:
@adamcrume Seems like this issue is no longer open with the recent PR merge?

I was searching for `Good First Issue` and only this one came up :-(

adamcrume:
Yep, this was fixed by https://github.com/tensorflow/rust/pull/263, and the issue didn't get closed somehow.  Sorry about that.  I'll go ahead and close it.  I filed some new issues from TODOs in the code and labeled them with `good first issue`.  #289 is probably the easiest, followed by #288.

PR #263, opened 2020-06-23
Wrap TF_GetOpList [#166]

kykosic:
I'm opening this as a draft to get some feedback, as I'm kinda stuck

Issues:
* Is this Library struct member op_list the correct place/type to store the proto?
* I'm having issues testing this for correctness. So far, I've tried loading the test library defined for the TF C-API unit test `test_op1.so` as compiled by bazel (https://github.com/tensorflow/tensorflow/blob/master/tensorflow/c/c_api_test.cc#L196-L212), but the resulting TF_Buffer always has a size of 0. I believe this test object should have a single OpDef in it. Is there a way we are implementing unit tests for this operation since it has a compiled C++ dependency?

Any advice would be appreciated.

EDIT: This is hopefully ready for review. See comments.


src/lib.rs
```diff
@@ -1488,14 +1489,23 @@ impl Library {
         let c_filename = CString::new(library_filename)?;
         let mut status = Status::new();
         let inner = unsafe { tf::TF_LoadLibrary(c_filename.as_ptr(), status.inner()) };
+
+        let buf = unsafe {
+            let mut buf = tf::TF_GetOpList(inner);
```
adamcrume, line 1494: `TF_GetOpList` cannot be called with NULL, so `inner` needs to be checked first.

src/lib.rs
```diff
@@ -1488,14 +1489,23 @@ impl Library {
...
+            Buffer::<u8>::from_c(&mut buf, false)
```
adamcrume, line 1495: The C API changed the convention for `TF_Buffer` from being heap-allocated to stack-allocated, which makes `Buffer` awkward to use with `TF_GetOpList`.  I think we need something like:

```rust
let buf = unsafe {
    let buf = tf::TF_GetOpList(inner);
    let heap_buf = tf::TF_NewBuffer();
    *heap_buf = buf;
    Buffer::<u8>::from_c(heap_buf, true)
};
```

src/lib.rs
kykosic: I've moved the `TF_GetOpList` call to after the `inner.null()` check

src/lib.rs
kykosic: Thanks for this, did not realize the `TF_Buffer` was stack-allocated. I've implemented your suggested method, however I needed to use `owned: false` in `::from_c`, otherwise there are memory address errors.

kykosic:
@adamcrume I

> Storing the op_list in the Library struct is fine.
> 
> I don't know if the comments I left will help with the test issue you were seeing, but they might.
> 
> I don't quite understand your last question; are you asking if we have prior art for how to write this test? We don't, unfortunately. We'll probably need to check in a generated .so and only run the tests in Linux, and a Windows user could do the same with a Windows library. Building the .so from source every time wouldn't be feasible since we need to support builds where the TensorFlow library is downloaded, which doesn't include the test .so.

I've added a test library at `test_resources/library/test_op.cc` that contains a single Op for testing. I've pre-compiled it on MacOS and Linux and included those objects in `test_resources/library/<os>/test_op.so`. The unit test will only trigger on these two OS at the moment; I don't know how to generate this test for Windows.

src/lib.rs
```diff
@@ -1480,6 +1480,7 @@ impl<T: TensorType> Display for Tensor<T> {
 #[derive(Debug)]
 pub struct Library {
     inner: *mut tf::TF_Library,
+    op_list: protos::op_def::OpList,
```
adamcrume, line 1483: Since this field is private (which is a good thing), there's no way for clients to get the list.  `Library` will need an accessor for clients to get the data, and the test should test against that.

kykosic: Sorry for the delay. Would this accessor want to return the proto object? Or do we want to wrap the OpList (and internal OpDefs) in some new struct to avoid exposing protos?

adamcrume: I've been avoiding exposing the protos in the API directly so far, since that would force users to use a specific proto library.  It might still be worth doing at some point for the convenience, but I don't want to do that in the middle of this pull request, at least.  We'll need a new struct to expose the information we need.  We can start with just the basics and add fields later if necessary.

kykosic: I've started by adding an `OpList` and `OpDef` struct to abstract the protos. I tried to follow patterns used elsewhere in the crate, but I want to make sure the design I'm following makes sense before implementing the `OpDef_ArgDef` and `OpDef_AttrDef` structs as well.

kykosic:
I've fixed the suggested changes, and added the `OpArgDef` and `OpAttrDef` structs. My remaining concerns for this pull:

1. I had to put `Library::load` behind the "experimental_training" feature flag, because the internal `OpArgDef::from_proto` requires `DataType::from_proto`, which is behind this feature flag.
2. Line 1501 `Buffer::<u8>::from_c(heaf_buf, false)` I not been able to get rid of `owned: false` argument. Setting this to `true` causes frequent segfaults when deallocating the TF_Buffer. I have not been able to remedy this or effectively debug it...
3. I'm not sure how the library wants to expose the proto `AttrValue` (fields `OpAttrDef.default_value` and `OpAttrDef.allowed_values`, and I have thus far omitted them. This `AttrValue` proto is some monstrosity enum of many other types, including `TensorProto`.

kykosic:
Thanks for the help @adamcrume!

adamcrume:
Thanks a lot for the feature work!

PR #268, opened 2020-08-03
Feature: ndarray interop with tensors [#260]

### LINKED ISSUE #260: Tensor interop with `ndarray`
I've found working with raw Tensor objects in rust to be somewhat tedious. I think it would be useful to add methods to cast between `Tensor` structs and [ndarray objects](https://docs.rs/ndarray/0.13.1/ndarray/) similar to how core TensorFlow python has builtin methods for casting with numpy. 

The Tensor object could impl directly a `Tensor::from_ndarray` method and `Tensor.to_ndarray` method to support this. Alternatively this could be made generic using `std::convert::TryFrom` and `std::convert::TryInto`, and implementations can be provided for correct ndarray objects.

Any thoughts, comments, or concerns about this idea?

kykosic:
Introduces an optional feature "ndarray" which provides two `From` impls:
* `ndarray::ArrayBase<S, D>` -> `tensorflow:Tensor<S>` – This allows converting from the various array types in "ndarray" to a Tensor with syntax:
```rust
 let tensor = Tensor::from(array);
```
* `tensorflow::Tensor<T>` -> `ndarray::Array<T, Dim<IxDyn>>` – This allows converting from a Tensor to an "ndarray" Array with owned data and dynamic dimensionality. (The dynamic dimensionality is necessary as Arrays have additional types for dimensions and is the only way to satisfy the `From` trait.) The conversion can be performed with:
```rust
let array = Array::from(tensor);
```

Converting to arrays without dynamic dimensionality can be performed by the user with [into_dimensionality](https://docs.rs/ndarray/0.13.1/ndarray/struct.ArrayBase.html#method.into_dimensionality), otherwise we would have to use multiple `TryFrom` traits for each dimension type or implement some conversion macro.

Closes #260


src/lib.rs
```diff
@@ -2240,6 +2276,62 @@ mod tests {
...
+        let expected = Tensor::new(&[2, 2]).with_values(&[0, 0, 0, 0]).unwrap();
+        assert_eq!(expected, tensor);
+
+        let arr = Array::<u8, _>::zeros((3, 2));
+        let tensor = Tensor::from(arr);
+        let expected = Tensor::new(&[3, 2])
+            .with_values(&[0, 0, 0, 0, 0, 0])
+            .unwrap();
+        assert_eq!(expected, tensor);
+    }
```
adamcrume, line 2303: This could use a test with non-default values so we can verify that the tensor values are actually being set.

src/lib.rs
```diff
@@ -2240,6 +2276,62 @@ mod tests {
...
+        let expected = Array::<i32, _>::zeros((2, 2)).into_dyn();
+        assert_eq!(expected, arr);
+
+        let tensor = Tensor::<u8>::new(&[3, 2])
+            .with_values(&[0, 0, 0, 0, 0, 0])
+            .unwrap();
+        let arr = Array::from(tensor);
+        let expected = Array::<u8, _>::zeros((3, 2)).into_dyn();
+        assert_eq!(expected, arr);
+    }
```
adamcrume, line 2333: This needs some non-default values, too.

kykosic:
I made the tests into a macro so they wouldn't be so repetitive and added cases with non-default values.

adamcrume:
Thanks!

PR #273, opened 2020-09-08
Add back std:: to io

Dispersia:
Hello, https://github.com/tensorflow/rust/commit/713daef09a22b47ae97d529b8ae43b6866291df5 removed `self` from io, however it was used under this compilation target. I just added it to the target itself instead of back at the top so it doesn't give unused warnings when not msvc


adamcrume:
Whoops, conditional compilation makes things tricky.  Thanks!

Issue #289, opened 2021-01-04
Test set_bool_list/get_bool_list


adamcrume:
In the `operation_attributes` test in graph.rs, `set_bool_list` and `get_bool_list` should be tested.

whereistejas:
Hi, @adamcrume! I'm working on this issue. Could you please assign it to me? I'm a newb at this, but I have started writing the code.

whereistejas:
Am I required to sign the CLA before the issue is assigned to me?

adamcrume:
You'll have to sign the CLA before I review the pull request. You can go ahead and send the pull request whenever you're ready, and the CLA bot will tell you if it checks out or not.

Heidar-An:
is this available? I'd be interested in having a go at it.

adamcrume:
https://github.com/tensorflow/rust/pull/291 hasn't been updated in a couple of years, so I'd say this is available.  I mentioned in https://github.com/tensorflow/rust/pull/291#pullrequestreview-564013073 that it's actually a bit difficult:

> Whoops, I'm sorry, I forgot this is actually difficult to fix at the moment. The test won't pass unless we use a real op and a real attribute, and I missed a comment in the code that mentions that there aren't any built-in ops with list(bool) attributes (which I just double-checked).
>
> We could create a custom op and load it, but that can be difficult, because it has to be written in C++, and linking that in is tricky.

You're welcome to work on this, but I'm going to remove the "good first issue" label.

Heidar-An:
It sounds pretty complex, so I might skip this for now (hopefully, I can come back to it a few months down the line :)) is there anything a beginner could start helping with? even something like documentation or something else introductory / not too hard to do, just so I can get a taste of the codebase.

PR #291, opened 2021-01-07
Test for get_bool_list/set_bool_list

whereistejas:
Hi,

This is my first ever contribution to a rust project. Please, let me know if the code is lacking in any way.

While setting the boolean list attribute, I couldn't find/think of any attribute with a type of boolean list. So, right now, it is just given  a random name `attr1`. Could you please suggest an appropriate attribute to use in its place?

Issue #289

Signed-off-by: Tejas Sanap <sanap.tejas@gmail.com>

### DIFF
--- a/src/graph.rs
+++ b/src/graph.rs
@@ -2628,6 +2628,17 @@ mod tests {
         );
         assert_eq!(42, op.get_attr_int("seed").unwrap());
 
+        // Issue #289
+        let op = {
+            let mut nd = g.new_operation("CheckBools", "CheckBools").unwrap();
+            nd.set_attr_bool_list("attr1", &[true, true]).unwrap();
+            nd.finish().unwrap()
+        };
+        assert_eq!(
+            &[true, true],
+            &op.get_attr_bool_list("attr1").unwrap() as &[bool]
+        );
+
         // TODO: Support get_attr_*/set_attr_*:
         // - bool_list
         // - tensor_list


adamcrume (commented):
Whoops, I'm sorry, I forgot this is actually difficult to fix at the moment.  The test won't pass unless we use a real op and a real attribute, and I missed a comment in the code that mentions that there aren't any built-in ops with `list(bool)` attributes (which I just double-checked).

We could create a custom op and load it, but that can be difficult, because it has to be written in C++, and linking that in is tricky.

whereistejas:
Is there any way to create the custom operation in Python? I know python well-enough. I have never really used C++.

whereistejas:
You mean we will have to do something like [this](https://www.tensorflow.org/guide/create_op) ?

adamcrume:
Yes, something like that would have to be done.  Creating a custom op and using it is currently tricky, because it requires compiling C++ and loading a shared library.  We do that for a different test elsewhere, but I would prefer a better implementation, because currently we're compiling the library manually and checking in a binary, which makes it difficult to make changes, is less safe, and causes cross-platform issues.

I'd prefer to do something like use https://crates.io/crates/cpp, but we'd have to make sure the TensorFlow headers are available, which likely requires adding it as a Git submodule.

whereistejas:
I have very limited knowledge in terms of tensorflow, but is there no way to bypass the use of C++, at all?

adamcrume:
As far as I'm aware, new ops can only be defined in C++.  The Rust bindings rely on the C API (https://github.com/tensorflow/tensorflow/blob/master/tensorflow/c/c_api.h), and I don't see anything in there for defining new ops.

PR #292, opened 2021-01-11
upgrade to tensorflow 2.4.0

RCasatta:
I've got a model built and saved with preprocessing layers on tensorflow 2.4.0 that fails to load with

```
Error: {inner:0x7fc3bfcf3da0, NotFound: Op type not registered 'DenseBincount' in binary running on <mypc>. Make sure the Op and Kernel are registered in the binary running in this process. Note that if you are loading a saved graph which used ops from tf.contrib, accessing (e.g.) `tf.contrib.resampler` should be done before importing the graph, as contrib ops are lazily registered when the module is first accessed.}
```

Upgrading the version in the build.rs file fixes it

### DIFF
--- a/tensorflow-sys/build.rs
+++ b/tensorflow-sys/build.rs
@@ -24,8 +24,8 @@ const REPOSITORY: &'static str = "https://github.com/tensorflow/tensorflow.git";
 const FRAMEWORK_TARGET: &'static str = "tensorflow:libtensorflow_framework";
 const TARGET: &'static str = "tensorflow:libtensorflow";
 // `VERSION` and `TAG` are separate because the tag is not always `'v' + VERSION`.
-const VERSION: &'static str = "2.3.0";
-const TAG: &'static str = "v2.3.0";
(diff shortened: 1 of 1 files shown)


adamcrume:
Please ensure that the `test-all` script passes.  Upgrading the TensorFlow version requires changing the version number in a few more places.

RCasatta:
I see... After updating a couple of other places I see tests are not passing because

"C-API functions TF_StringDecode, TF_StringEncode, and TF_StringEncodedSize are no longer relevant and have been removed; see core/platform/ctstring.h for string access/modification in C."

I'll give it a try but I think I can't easily tackle the changes needed in `impl TensorType for String`

adamcrume:
I'm trying to upgrade it myself, but besides the obvious breaking changes, they're also using inline functions in the C API which would require the Rust bindings to compile and link in their own C code (which we're not currently doing), and I haven't tackled that can of worms, yet.

adamcrume:
Closing as obsolete since we're now on 2.6.

Issue #290, opened 2021-01-04
Add OpArgDef::default_value and allowed_values


adamcrume:
This will require adding a new enum which looks like the generated `protos::AttrValue_oneof_value` (in protos/attr_value.rs), and  `default_value` will need to be optional.

whereistejas:
Hi @adamcrume! Do you think I can take up this issue instead of #289 ? Could you please give me some pointers on how to get started with this one?

adamcrume:
Yeah, you can definitely take this.  This needs a new enum in src/lib.rs (probably called `AttrValue`) which looks like `AttrValue_oneof_value` in src/protos/attr_value.rs.  Some of the types already have non-proto equivalents, like `tensorflow::Shape` instead of `tensorflow::protos::tensor_shape::TensorShapeProto`.  It looks like we need a `NameAttrList` struct as well.  These will need conversions to/from protos, and you can see how we've done others like `OpDef`.

whereistejas:
Hi @adamcrume,
I thought I would outline the changes I'm going to make in a comment before actually starting to code.

There is a struct called `OpArgDef`:
https://github.com/tensorflow/rust/blob/ceefb76ad8aa7b7e2fe0724279fead86ffe35539/src/lib.rs#L1712-L1723

To fill the fields `default_value` and `allowed_values` in `OpArgDef` we need a new enum called `AttrValue`.
This `AttrValue` enum looks like the enum `AttrValue_oneof_value`:
https://github.com/tensorflow/rust/blob/ceefb76ad8aa7b7e2fe0724279fead86ffe35539/src/protos/attr_value.rs#L42-L53

**Do I need add/remove any variants from this enum to in `AttrValue` or can I take it as it is? Where do I need to create this enum?
Should I create it in `src/lib.rs`? The reason I'm asking this is, I saw that `attr_value.rs` is a generated file, so can we edit it?**

Here, the two new fields are from the struct `OpDef_AttrDef`:
https://github.com/tensorflow/rust/blob/ceefb76ad8aa7b7e2fe0724279fead86ffe35539/src/protos/op_def.rs#L1176-L1188

adamcrume:
[...]

PR #293, opened 2021-01-26
Adding AttrValue enum

whereistejas:
Hi @adamcrume,

I have opened this PR. I thought we could move our discussion from the issue page to here.

TODO:
Struct definitions for AttrValue::list and AttrValue::func are still
pending

Resolves issue: #290

Signed-off-by: Tejas Sanap <sanap.tejas@gmail.com>

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -1721,6 +1721,8 @@ pub struct OpArgDef {
     type_list_attr: String,
     is_ref: bool,
     // TODO: Add "default_value" and "allowed_values" from OpDef_AttrDef proto
+    default_value: Option<AttrValue>,
+    allowed_values: Option<AttrValue>,
 }
 
 impl OpArgDef {
@@ -1848,10 +1850,10 @@ pub enum AttrValue {
     I(i64), 
     F(f32), 
     B(bool), 
-    Field_type(protos::types::DataType), 
+    Field_type(DataType), 
     Shape(Shape), 
     Tensor(protos::tensor::TensorProto), 
-    List(AttrValue_ListValue), 
+    List(AttrValueList), 
     Func(NameAttrList), 
     Placeholder(String),
 }
@@ -1863,13 +1865,10 @@ pub struct NameAttrList {
     // TODO: the AttrValue type in this hashmap is the incorrect type. Please ref to
     // protos/attr_value.rs:1381
     pub attr: HashMap<::std::string::String, AttrValue>,
-    // special fields
-    pub unknown_fields: UnknownFields,
-    pub cached_size: CachedSize,
 }
 
 #[derive(Debug)]
-pub struct AttrValue_ListValue {}
+pub struct AttrValueList {}
 
 ////////////////////////
 
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -20,7 +20,8 @@ use libc::{c_int, c_uint};
 #[cfg(feature = "ndarray")]
 use ndarray::{Array, ArrayBase, Data, Dim, Dimension, IxDynImpl};
 use num_complex::Complex;
-use protobuf::ProtobufEnum;
+use protobuf::{ProtobufEnum, CachedSize, UnknownFields};
+use std::collections::HashMap;
 use std::borrow::Borrow;
 use std::cell::Cell;
 use std::cell::RefCell;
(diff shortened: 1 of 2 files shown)


whereistejas:
I have a couple of questions.
1. In the struct `NameAttrList` there is a field called `attr`. We can see that it's type is:
https://github.com/tensorflow/rust/blob/ceefb76ad8aa7b7e2fe0724279fead86ffe35539/src/protos/attr_value.rs#L1377-L1385
The type of the hashmap's value itself is:
https://github.com/tensorflow/rust/blob/ceefb76ad8aa7b7e2fe0724279fead86ffe35539/src/protos/attr_value.rs#L26-L33
Do we have an equivalent of the `AttrValue` in `protos/attr_value.rs` in `src/lib.rs` the way we have for `shape`? I couldn't find any.
2. I'm curious about how the enums and structs we are creating fits into the bigger picture. Why are we exactly adding this `AttrValue` enum in lib.rs? What will it enable us to do? What problems will it solve? I know this question sounds kind of stupid, but I want to understand the functional side of the code we are writing.

src/lib.rs
```diff
@@ -1839,6 +1840,37 @@ impl OpAttrDef {
...
 
+/// Types of attribute values for an operation
+#[non_exhaustive]
+#[derive(Debug)]
+pub enum AttrValue {
+    S(Vec<u8>), 
+    I(i64), 
+    F(f32), 
+    B(bool), 
+    Field_type(protos::types::DataType),
```
adamcrume, line 1851: This should be the native Rust enum, not the protobuf enum (i.e. `tensorflow::DataType`, not `protos::types::DataType`).

src/lib.rs
```diff
@@ -1839,6 +1840,37 @@ impl OpAttrDef {
...
+    // TODO: the AttrValue type in this hashmap is the incorrect type. Please ref to
+    // protos/attr_value.rs:1381
+    pub attr: HashMap<::std::string::String, AttrValue>,
+    // special fields
+    pub unknown_fields: UnknownFields,
+    pub cached_size: CachedSize,
+}
+
+#[derive(Debug)]
+pub struct AttrValue_ListValue {}
```
adamcrume, line 1872: This doesn't need to use exactly the same name as the proto, especially since the generated names are a weird mix of CamelCase and snake_case.  This likely makes more sense as either `AttrValueList` or `AttrListValue`.

adamcrume (commented):
`protos::AttrValue`/`protos::AttrValue_oneof_value` is a workaround for a limitation of protos, which is that a repeated field can't be part of a oneof, so people often create a message which only holds a repeated field, and then that message can be part of the oneof.  Essentially, we can inline `AttrValue_oneof_value`, i.e. `tensorflow::AttrValue` would be an enum consisting of the possible values in `AttrValue_oneof_value`, which is what you've already done.

The bigger picture here (the issue this PR is addressing) is that we want to support `OpArgDef::default_value` and `allowed_values` but with the more ergonomic Rust-native types instead of the generated proto code, which can be harder to deal with.

src/lib.rs
```diff
@@ -1720,6 +1721,8 @@ pub struct OpArgDef {
     type_list_attr: String,
     is_ref: bool,
     // TODO: Add "default_value" and "allowed_values" from OpDef_AttrDef proto
+    default_value: Option<AttrValue>,
+    allowed_values: Option<AttrValue>,
```
whereistejas, line 1725: Hi @adamcrume 
Thank you for your helpful comments.

> The bigger picture here (the issue this PR is addressing) is that we want to support `OpArgDef::default_value` and `allowed_values` but with the more ergonomic Rust-native types instead of the generated proto code, which can be harder to deal with.

So, if I understand this correctly, we want the fields `allowed_values` and `default_values` in `OpDefArg` to be of type `Option<AttrValue>`. However, in `src/lib.rs` `AttrValue` is an enum unlike in protos where it is a struct.

src/lib.rs
```diff
@@ -1839,6 +1842,43 @@ impl OpAttrDef {
...
+#[non_exhaustive]
+#[derive(Debug)]
+pub enum AttrValue {
+    S(Vec<u8>), 
+    I(i64), 
+    F(f32), 
+    B(bool), 
+    Field_type(DataType), 
+    Shape(Shape), 
+    Tensor(protos::tensor::TensorProto),
```
whereistejas, line 1855: The type of the `Tensor` variant is also something we can change. There is a `Tensor` type in `src/lib.rs` but it requires us to mention a datatype when defining it.

https://github.com/tensorflow/rust/blob/ceefb76ad8aa7b7e2fe0724279fead86ffe35539/src/lib.rs#L1257-L1260

adamcrume: I meant to mention `AnyTensor`, but it looks like I forgot.  We will probably need to make it public and then this can be `Box<dyn AnyTensor>`.

src/lib.rs
adamcrume: Structs with `Option` fields can still derive `Clone`, `PartialEq`, and `Eq`.  Can you explain what you mean?  It does require that `AttrValue` implement those traits, though.

whereistejas: When adding the `Clone`, `PartialEq`, and `Eq` traits to the structure the `NameAttrList`, I'm required to apply the same traits to the enum `AttrValue`, but it has fields of type `f32` and `Box` which don't have these traits. I've been stuck here, for a while, now. I will add the cargo check log in a separate gist [here](https://gist.github.com/whereistejas/2da739363c5410ece5414d4f7a3a954e). Am I missing some Rust fundamental about how to assign these traits? Because this becomes an infinite cycle where have I to assign traits to the structs within these structs and the structs within them and so on....

PR #294, opened 2021-02-01
Add Scope::with_kernel_label

siyavash:
Implement with_kernel_label for Scope including:

- Add kernel_label field to Scope.
- Set the field with Scope::with_kernel_label.
- Update new_operation to set the corresponding attr.

### DIFF
--- a/src/scope.rs
+++ b/src/scope.rs
@@ -162,7 +162,7 @@ impl Scope {
             },
             device: self.device.clone(),
             control_deps: self.control_deps.clone(),
-            kernel_label: "".to_string(),
+            kernel_label: self.kernel_label.clone(),
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/scope.rs
```diff
@@ -160,6 +162,7 @@ impl Scope {
             },
             device: self.device.clone(),
             control_deps: self.control_deps.clone(),
+            kernel_label: "".to_string(),
```
adamcrume, line 165: The `kernel_label` should be inherited from the parent scope (i.e. `kernel_label: self.kernel_label.clone()`) here and in the other methods below.

adamcrume (changes requested):

src/scope.rs
siyavash: Done.

adamcrume:
Thanks!

PR #295, opened 2021-02-10
Implement Scope::with_xla_cluster

siyavash:
Implement with_xla_cluster for Scope including:

- Add xla_cluster field to Scope.
- Set the field with Scope::with_xla_cluster.
- Update new_operation to set the `_XlaCluster` attr.


adamcrume:
Thanks!

PR #299, opened 2021-05-11
Add dtype and shape accessors to TensorInfo

### LINKED ISSUE #298: Expose shape and data type info in TensorInfo
It would be nice to be able to get the [shape and data type from the TensorInfo struct](https://github.com/tensorflow/rust/blob/e51522804cb329dd46065a526d892ece66c8dc4d/src/saved_model.rs#L111) so you can load a model and get that information.

bjchambers:
Fixes #298.

### DIFF
--- a/src/saved_model.rs
+++ b/src/saved_model.rs
@@ -125,6 +125,16 @@ impl TensorInfo {
         &self.name
     }
 
+    /// Returns the data type of the tensor.
+    pub fn dtype(&self) -> DataType {
+        self.dtype
+    }
+
+    /// Returns the shape of the tensor.
+    pub fn shape(&self) -> &Shape {
+        &self.shape
+    }
+
     // We don't use Into, because we don't want this to be public API.
     fn into_proto(self) -> protos::meta_graph::TensorInfo {
         let mut info = protos::meta_graph::TensorInfo::new();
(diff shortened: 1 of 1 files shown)


src/saved_model.rs
```diff
@@ -124,6 +124,16 @@ impl TensorInfo {
     pub fn name(&self) -> &OutputName {
         &self.name
     }
+
```
adamcrume, line 127: Please remove the whitespace on these blank lines.

src/saved_model.rs
```diff
@@ -124,6 +124,16 @@ impl TensorInfo {
     pub fn name(&self) -> &OutputName {
         &self.name
     }
+    
+    /// Returns the data type of the tensor.
+    pub fn dtype(&self) -> &DataType {
```
adamcrume, line 129: Since `DataType` is trivial to copy (i.e. is small and implements `Copy`), we can probably go ahead and just return it by value.

adamcrume (changes requested):

src/saved_model.rs
bjchambers: Done.

src/saved_model.rs
bjchambers: Whoops. Good idea. Done.

adamcrume:
Thanks!

PR #307, opened 2021-07-10
Update addition example's python code for tf2.5

dskkato:
I tried to migrate Python code using TensorFlow 2.x version.

In TensorFlow 2.x, the eager mode is ON by default, which makes it hard to get a concrete computation graph. One way to obtain a similar result is to wrap the function and use `get_concrete_function` method like in this PR.

Please refer to TF documents for detail : https://www.tensorflow.org/guide/function

~~I also tried to change outputs node name (which is `Identity` now), for compatibility, but could not find any solution regarding to this point.~~ Please note that this PR makes the `addition` example totally incompatible with TensorFlow 1.x versions.

I'm glad to take any feedback. And if you already working on this point in a different way, please let it go.


dskkato:
@googlebot I signed it!

adamcrume:
Thanks for the update!  This looks good to me.  I think it's fine if the examples require TensorFlow 2, since that's what the library is now targeting.

ptitfred:
Please note that when ran as advertised (in the `examples/addition` directory) it creates another subtree `examples/addition/examples/addition`.

PR #310, opened 2021-07-18
add pre-trained Keras model use example

dskkato:
This example demonstrates how to use Keras model saved as  TensorFlow's SavedModel format.

Currently, this does not contain any model files (like .pb files) because of their file size.


examples/mobilenetv3/README.md
```diff
@@ -0,0 +1,23 @@
+## About sample image
+
+A sample image (image file in `218`) was obtained from the ImageNet dataset.
```
adamcrume, line 3: I don't think we can distribute ImageNet images.  The ImageNet terms and conditions state that the data can only be shared with people who agree to the terms and conditions first, and we can't guarantee that.  This will have to be replaced with an image that we have the right to distribute under this project's license (Apache 2.0). If you can provide a photo that you take yourself of a non-copyrighted subject, I think that should work.

examples/mobilenetv3/README.md
```diff
@@ -0,0 +1,23 @@
...
+
+## About model inputs/outputs nodes
+
+There are several ways to verify model inputs/outputs nodes. One convinient way to interact with SavedModel format is `saved_model_cli` which is installed with tensorflow 2.x.
```
adamcrume, line 7: Typo: convenient

adamcrume (changes requested):

examples/mobilenetv3/README.md
dskkato: Now fixed.

examples/mobilenetv3.rs
```diff
@@ -0,0 +1,84 @@
...
+
+    // Run the graph.
+    let mut args = SessionRunArgs::new();
+    args.add_feed(op_x, 0, &x);
+    let token_output = args.request_fetch(op_output, 0);
+    session.run(&mut args)?;
+
+    // Check our results.
+    let output: Tensor<f32> = args.fetch(token_output)?;
+    let mut res = Vec::<f32>::new();
```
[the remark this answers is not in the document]
dskkato: Good to know. Thanks!

examples/mobilenetv3.rs
```diff
@@ -0,0 +1,84 @@
...
+use tensorflow::Status;
+use tensorflow::Tensor;
+use tensorflow::DEFAULT_SERVING_SIGNATURE_DEF_KEY;
+
+use image::io::Reader as ImageReader;
+use image::GenericImageView;
+
+fn main() -> Result<(), Box<dyn Error>> {
+    let export_dir = "examples/mobilenetv3";
+    if !Path::new(export_dir).exists() {
```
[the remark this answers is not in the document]
dskkato: The code now checks the saved_model.pb as suggested. Thanks.

examples/mobilenetv3/README.md
dskkato: The photo was replaced to that I've taken. Could you check this again?

adamcrume:
Thanks!

Issue #309, opened 2021-07-16
Migration of Python examples to TF2.5


dskkato:
In a version 0.17.0 release, the back end of this crate became TF 2.5, while following example Python codes stay for TF 1.x serieas.

* [x] addition - resolved by #307
* [ ] regression
* [ ] regression_checkpoint
* [ ] regression_savedmodel

In addition to those examples, I also would like to add some Keras examples like #286 since its signature usage was helpful for me.

Would you mind if I try these migrations? I already confirmed basic TF2.5 usages posted in [my repository](https://github.com/dskkato/tf2_python_rust).

adamcrume:
By all means, updating the Python to TF 2.5 is helpful and examples are always welcome!

PR #312, opened 2021-07-30
Update regression_savedmodel example for tf2.5

dskkato:
In this PR, I tried to update `regression_savedmodel` example targetting TF 2.x. Signature handling becomes a bit verbose, but you can find raw operation names like `StatefulPartitionedCall_2:0` in a readable manner.

This PR is a part of #309.


adamcrume:
Thanks!

PR #313, opened 2021-08-10
Fix op-codegen build error

dskkato:
op-codegen seems to be broken due to nom's API changes, though I'm not sure about this tool in detail.


adamcrume:
Thanks!

PR #320, opened 2021-09-07
remove unused internal crate (libtensorflow-sys)

dskkato:
This PR is just for clean up the internal binding crate since it was migrated to `tensorflow-sys` from `libtensorflow-sys` as is in #6.


adamcrume:
Thanks!

Issue #321, opened 2021-09-08
Eager API support


dskkato:
_This is a repost from the mailing list to track activities._

-----

I've been using TensorFlow and Rust for a while now, and I'm very happy with this binding. Since this crate implements many APIs other than inference, I wanted to have Eager API as well. As far as I know, cppflow has partial support for the Eager API op, so I think I can make a basic implementation by referring to that and tensorflow's eager/c_api_test.

I think there are three major tasks required for this implementation.

* [x] Creating bindings in `tensorflow-sys`.
* [x] Implementing the Eager API
* [x] Implementation of raw op (refer to cppflow)

At present, I believe that I can implement an API that can handle both Tensor and TensorHandle relatively transparently by leaving the Tensor struct as it is and implementing a common trait between Tensor and the newly implemented TensorHandle.

I myself often use TensorFlow for image-related tasks, and the Eager
API will be useful for basic I/O without relying on image crate.

If you have any suggestions or comments, I would appreciate your feedback.

Thank you.

dskkato:
The Eager API is not guaranteed as stable as `tensorflow/c/c_api.h`, so it might be better not to include it as default functions.

```
// C API extensions to experiment with eager execution of kernels.
// WARNING: Unlike tensorflow/c/c_api.h, the API here is not guaranteed to be
// stable and can change without notice.
```

dskkato:
@adamcrume Up to #329 and #344 are the implementation of the eager API that I had originally envisioned. I think users should be able to try out the eager API now.

dskkato:
Close since all relevant patches were merged.

PR #322, opened 2021-09-08
Include eager header

dskkato:
I updated the binding part of `tensorflow-sys` so that we could implement the eager API later on.

This PR is related to #321.


dskkato:
I will close this PR for now as I'm working on the configuration of the Eager API.

PR #324, opened 2021-09-22
[codegen] update codegens

dskkato:
This patch enables to parse ops.pbtxt in TF2.6 without errors.

Some code for TF operations are generated by parsing `ops.pbtxt`. but it fails to parse it in TF2.6. It seems to be due to parse error arising with `is_distributed_communication` and `experimental_full_type` nodes. To resolve this process, I have made the following fixes:

1. add "full_type.proto" in the `codegen-proto` project.
  This type is required to parse `experimental_full_type` node.
2. re-generate proto code.
3. add some parser to deal the above nodes in the `codegen-op` project.
4. re-generate `ops_impl.rs`.

I hope this helps.


tensorflow-op-codegen/src/parser.rs
```diff
@@ -155,6 +157,40 @@ fn data_type<'a>(input: &'a [u8]) -> ParseResult<'a, DataType> {
     })(input)
 }
 
+fn type_id<'a>(input: &'a [u8]) -> ParseResult<'a, FullTypeId> {
+    map_res(identifier, |s| match &s as &str {
```
adamcrume, line 161: I know this is how we do it above, but this would be more robust if parsed with something like:

```rust
map_res(identifier, |s| {
    match FullTypeId::values()
        .iter()
        .filter(|d| d.descriptor().name() == s)
        .next()
    {
        Some(d) => Ok(FullTypeId::from_i32(d.value()).unwrap()),
        None => Err(nom::Err::<VerboseError<&'a [u8]>>::Error(make_error(
            input,
            ErrorKind::Alt,
        ))),
    }
})(input)
```

That way, if new enum values are added in the future, this code doesn't need to be updated.

(Ideally this would use `EnumDescriptor::value_by_name`, but that panics on error rather than returning a `Result` or `Option` for some reason.)

adamcrume (changes requested):

tensorflow-op-codegen/src/parser.rs
dskkato: Thanks for the review. I have changed the code to the one you suggested.

Shall I change `fn data_type` as well?

PR #326, opened 2021-10-04
Eager API support

### LINKED ISSUE #321: Eager API support

dskkato:
resolves #321


src/eager/mod.rs
```diff
@@ -0,0 +1,835 @@
...
+
+struct Op {
+    inner: *mut tf::TFE_Op,
+}
+
+impl Op {
+    fn new(ctx: &Context, op_or_function_name: &str) -> Result<Op> {
+        let status = Status::new();
+
+        let op_or_function_name = CString::new(op_or_function_name).unwrap();
```
adamcrume, line 381: These should avoid `unwrap()` in general.

src/eager/mod.rs
```diff
@@ -0,0 +1,835 @@
...
+    }
+}
+
+impl Drop for Op {
+    fn drop(&mut self) {
+        unsafe { tf::TFE_DeleteOp(self.inner) };
+    }
+}
+
+///
```
adamcrume, line 659: This needs docs.

src/eager/mod.rs
```diff
@@ -0,0 +1,835 @@
...
+        TensorHandle::new(ctx, self)
+    }
+}
+
+impl<'a> ToHandle<'a> for TensorHandle<'a> {
+    fn to_handle(&self, _: &'a Context) -> Result<TensorHandle<'a>> {
+        self.copy_sharing_tensor()
+    }
+}
+#[cfg(test)]
```
adamcrume, line 679: Nit: need a blank line above this line.

src/eager/mod.rs
```diff
@@ -0,0 +1,835 @@
...
+        let opts = ContextOptions::new();
+        let ctx = Context::new(opts).unwrap();
+
+        let mut x = Tensor::new(&[1]);
+        x[0] = 2i32;
+
+        let x_handle = x.to_handle(&ctx).unwrap();
+
+        let y = unsafe {
+            let status = Status::new();
```
adamcrume, line 696: This status is never checked.

src/eager/mod.rs
```diff
@@ -0,0 +1,835 @@
...
+    fn add_test() {
+        let opts = ContextOptions::new();
+        let ctx = Context::new(opts).unwrap();
+
+        let x = Tensor::from(2);
+        let y = x.clone();
+
+        let h = raw_ops::add(&ctx, x, y).unwrap();
+        let z: Result<Tensor<i32>> = h.resolve();
+        assert!(z.is_ok());
```
adamcrume, line 714: The `unwrap()` would fail the test anyway if it weren't `ok`, so this assertion isn't really needed.

src/eager/mod.rs
```diff
@@ -0,0 +1,835 @@
...
+        let z = z.unwrap();
+        assert_eq!(z[0], 4i32);
+
+        let h = raw_ops::add(&ctx, z.clone(), z.clone()).unwrap();
+        let z: Tensor<i32> = h.resolve().unwrap();
+        assert_eq!(z[0], 8i32);
+
+        let h1 = z.clone().to_handle(&ctx).unwrap();
```
adamcrume, line 722: Why is `z` being cloned before calling `to_handle`?  `to_handle` doesn't consume the tensor, unless I'm misunderstanding something.

src/eager/mod.rs
```diff
@@ -0,0 +1,835 @@
...
+        let ctx = Context::new(opts).unwrap();
+
+        let filename: Tensor<String> =
+            Tensor::from(String::from("test_resources/io/sample_text.txt"));
+
+        let h = raw_ops::read_file(&ctx, filename).unwrap();
+        let z: Tensor<String> = h.resolve().unwrap();
+        assert_eq!(z.len(), 1);
+        assert_eq!(z[0].len(), 32);
+        assert_eq!(z[0], "This a sample text for unittest.")
```
adamcrume, line 751: These three assertions could be simplified to something like:

```
assert_eq!(&z[..], &["This is a sample text for unittest.".to_string()]);
```

src/eager/mod.rs
```diff
@@ -0,0 +1,835 @@
...
+    }
+
+    #[test]
+    fn context_test() {
+        let opts = ContextOptions::new();
+        let ctx = Context::new(opts).unwrap();
+
+        let devices = ctx.device_list().unwrap();
+        assert!(devices.len() > 0);
+        for d in devices.iter() {
```
adamcrume, line 815: `for d in &devices {` would be more idiomatic.

src/eager/mod.rs
```diff
@@ -0,0 +1,835 @@
...
+
+    #[test]
+    fn type_mismatch_test() {
+        let opts = ContextOptions::new();
+        let ctx = Context::new(opts).unwrap();
+
+        let t = Tensor::from(0);
+        let h = t.to_handle(&ctx).unwrap();
+        let v: Result<Tensor<f32>> = h.resolve();
+        assert!(v.is_err());
```
adamcrume, line 828: This is redundant with the `assert!(false)` below.

adamcrume (changes requested):
Sorry, I haven't been able to review everything, yet, but I figured I'd respond with the comments I had.  This is a very large change, and it'd be easier to review if broken up into smaller PRs.

dskkato:
Thank you for your so much feedbacks. I had to say that I am very sorry that I did not add an explanation in this pull request.

After checking your comments, I will break down this large PR into small pieces.

adamcrume:
I'd like to mention this PR in the contributor summit.  Are you fine with that?

dskkato:
Of course it's fine! Thank you.

src/eager/mod.rs
```diff
@@ -0,0 +1,835 @@
...
+///     let opts = ContextOptions::new();
+///     let ctx = Context::new(opts).unwrap();
+///
+///     let t = Tensor::from(3);
+///     let h = t.to_handle(&ctx).unwrap();
+///     h
+/// };
+/// ```
+#[derive(Debug)]
+pub struct TensorHandle<'a> {
```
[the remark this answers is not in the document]
dskkato: If TensorHandle were parameterized, all return types from raw operations also needed to be parameterized, but it was difficult to write such a generic code for me.

In more detail, the output tensor type  depends on input type for raw operation functions, which makes difficult to parameterize at compile time.

src/eager/mod.rs
dskkato: In order to avoid using unwrap, let's see if we can use CStr as the argument of new.

src/eager/mod.rs
dskkato: This is because I first write the ToHandle trait to consume the input Tensor, but switched not to do so later on, I think... I'll check it out.

src/eager/mod.rs
```diff
@@ -0,0 +1,835 @@
...
+        let mut status = Status::new();
+        unsafe {
+            tf::TFE_ContextOptionsSetConfig(
+                self.inner,
+                config.as_ptr() as *const _,
+                config.len(),
+                status.inner(),
+            );
+        }
+        if status.is_ok() {
```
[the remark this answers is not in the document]
dskkato: Fixed as you pointed out.

src/eager/mod.rs
```diff
@@ -0,0 +1,835 @@
...
+            })
+        }
+    }
+
+    /// Returns the DataType that corresponds to this type.
+    pub fn data_type(&self) -> DataType {
+        unsafe { DataType::from_c(tf::TFE_TensorHandleDataType(self.inner)) }
+    }
+
+    /// Return a number of dimensions.
```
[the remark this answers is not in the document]
dskkato: Fixed.

src/eager/mod.rs
```diff
@@ -0,0 +1,835 @@
...
+        let status = Status::new();
+        let num_dims = unsafe { tf::TFE_TensorHandleNumDims(self.inner, status.inner) };
+        if !status.is_ok() {
+            Err(status)
+        } else {
+            Ok(num_dims)
+        }
+    }
+
+    /// Return a number of elements
```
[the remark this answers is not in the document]
dskkato: Fixed.

src/eager/mod.rs
```diff
@@ -0,0 +1,835 @@
...
+        let status = Status::new();
+        let num_dims = unsafe { tf::TFE_TensorHandleNumElements(self.inner, status.inner) };
+        if !status.is_ok() {
+            Err(status)
+        } else {
+            Ok(num_dims)
+        }
+    }
+
+    /// Return a number elements for a givin dim_index.
```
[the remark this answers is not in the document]
dskkato: Fixed.

src/eager/mod.rs
```diff
@@ -0,0 +1,835 @@
...
+    /// device name is not always the device holding the tensor handle's memory. If
+    /// you want the latter, use TFE_TensorHandleBackingDeviceName. This function
+    /// will block till the operation that produces `h` has completed.
+    pub fn device_name(&self) -> Result<String> {
+        let status = Status::new();
+        unsafe {
+            let device_name = tf::TFE_TensorHandleDeviceName(self.inner, status.inner);
+            if status.is_ok() {
+                // todo: UTF8 check
+                Ok(CStr::from_ptr(device_name).to_str().unwrap().to_string())
```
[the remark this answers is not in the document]
dskkato: Fixed as you pointed out.

src/eager/mod.rs
dskkato: The status check code was added.

src/eager/mod.rs
dskkato: Fixed.

src/eager/mod.rs
dskkato: Fixed.

dskkato:
I think I've fixed most of the things that can be easily fixed, so I'm going to divide the PR into smaller parts based on the content here.

src/eager/mod.rs
```diff
@@ -0,0 +1,835 @@
...
+    }
+
+    /// Returns the DataType that corresponds to this type.
+    pub fn data_type(&self) -> DataType {
+        unsafe { DataType::from_c(tf::TFE_TensorHandleDataType(self.inner)) }
+    }
+
+    /// Return a number of dimensions.
+    /// This function will block till the operation that produces the TensorHandle has completed.
+    pub fn num_dims(&self) -> Result<i32> {
```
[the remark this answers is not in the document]
dskkato: I think I can leave the return type as is, since the following code uses the same type (while it is referred to as `c_int`). This code is in graph.rs.


```rust
    /// Returns the number of dimensions of the Tensor referenced by `output`.
    ///
    /// If the number of dimensions in the shape is unknown, returns -1.
    ///
    /// Returns an error if:
    ///
    ///   * `output` is not in `graph`.
    pub fn num_dims<I: Into<Output>>(&self, output: I) -> Result<c_int> {
```

dskkato:
Closes since this is progressing in the following smaller PRs:

* #328
* #331 
* #332 
* #338 
* #329

PR #316, opened 2021-08-20
Port CI tests to GitHub Actions

damienstanton:
This is a naive port of the current test suite over to GitHub Actions (GH Actions), and completes in 7-10 minutes. It is just a boilerplate runner that then executes the `test-all` script. Note that this GH Action does not replicate documentation generation and push. If that's a desired feature, please let me know and I'll update the PR to include it.

Fixes #301.

PR #327, opened 2021-11-09
Port CI tests to GitHub Actions --- Part 2

### LINKED ISSUE #301: Migrating from Travis CI to GitHub Actions?
Since travis-ci.org is shutting down, it probably makes sense to move this open project to another CI platform. GitHub actions seems like a low-friction environment.

Is this something the team would be interested in? If so, I would be willing to lead the effort to get a CI pipeline working again.

dskkato:
This PR tries to port CI tests to GitHub Actions. While #316 is for the same purpose, which is not active recently so I opened as a different PR.

The current target platforms in the configuration are Ubuntu, macOS, and Windows, with the latest stable channel Rust version. In addition, `.gitattribute` was modified to support the CI on the Windows env.

The clippy checking might be helpful, but it generates lots of warnings currently so they are commented out now.

resolves #301

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,5 +1,6 @@
 # Handle line endings automatically for files detected as text
 # and leave all files detected as binary untouched.
+*.rs text eol=lf
 * text=auto
 
 # Check for whitespace errors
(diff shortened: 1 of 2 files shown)


adamcrume:
Leaving clippy disabled for now makes sense.

I'm not sure how to test this other than just merging it in and seeing if it works, so let's do that.

dskkato:
Oops, the build status budge is still heading to TravisCI. I will fix it in another PR.

PR #328, opened 2021-11-10
[tensorflow-sys] Expose the eager API.

dskkato:
In this PR, the eager APIs are exposed via `tensorflow-sys`. The eager API is a very experimental stage, so it is separated as `eager` features.

The binding generation is done by using TF2.6. The original binding code is renamed from `bindgen.rs` to `c_api.rs`, and non-TF_* related symbols were removed.

This PR is related to #321 and #326.


tensorflow-sys/tests/lib.rs
```diff
@@ -8,3 +9,32 @@ fn linkage() {
...
     }
 }
+
+/// Test that the eager API works.
+#[cfg(feature = "eager")]
+#[test]
+fn tfe_tensor_handle() {
+    let data = vec![0.0f32; 100];
+    let shape = [1, 10, 10];
+    let num_elements = shape.iter().fold(1, |t, x| t * x) as usize;
```
adamcrume, line 19: This can be simplified to `shape.iter().product()`.

tensorflow-sys/tests/lib.rs
```diff
@@ -8,3 +9,32 @@ fn linkage() {
...
+            num_elements * std::mem::size_of::<f32>(),
+        );
+        std::ptr::copy_nonoverlapping(
+            data.as_ptr(),
+            ffi::TF_TensorData(tf_tensor) as *mut f32,
+            ffi::TF_TensorByteSize(tf_tensor) / std::mem::size_of::<f32>(),
+        );
+        let status = ffi::TF_NewStatus();
+        let tfe_handle = ffi::TFE_NewTensorHandle(tf_tensor, status);
+
```
adamcrume, line 35: This should check that the status is OK.

adamcrume (changes requested):

tensorflow-sys/tests/lib.rs
dskkato: Thanks. The code now uses `product`.

tensorflow-sys/tests/lib.rs
dskkato: The status check code is a copy of the `ok!` macro in the `multiplication.rs`.

adamcrume (approved):

adamcrume:
Thanks!

PR #329, opened 2021-11-10
[tensorflow-op-codegen] add codegen for eager API

dskkato:
This PR will add a code generation process of the `raw_ops` for the eager API.

* API information https://www.tensorflow.org/api_docs/python/tf/raw_ops
* ref: See [the RFC](https://github.com/tensorflow/community/blob/master/rfcs/20181225-tf-raw-ops.md) for details about the module.

The code generation process and the generated code are very similar to that for the graph construction, implemented in `src/ops/ops_impl.rs`. The differences are:

* `control_inputs` fields are removed since they do not make sense.
* `build` methods were renamed to `call` and updated to fit the eager execution context.

In addition to the above, here I have made some interface fixes:

* Some `number_attr`, `type_attr`, and `type_list_attr` that do not affect the execution results were removed.
* Some ops can now accept the array of tensors as arguments. See https://github.com/tensorflow/rust/issues/325#issuecomment-964418299

## Steps for the code generation

```
cd tensorflow-op-codegen; cargo run --bin eager -- $PATH_TO_TENSORFLOW $PWD/..
```

## Other information

This PR is related to #321 and #326, and was waiting for #338 to be resolved.


dskkato:
This PR is ready for review now. Mentioning it just in case @adamcrume missed it.

dskkato:
I’m afraid that this PR is needed to use the eager APIs otherwise there are no way to construct the eager operation... @adamcrume

tensorflow-op-codegen/src/bin/eager.rs
```diff
@@ -0,0 +1,783 @@
...
+                    .map(|x| format!("{}", x.get_size()))
+                    .collect();
+                if dims.is_empty() {
+                    write!(w, "None")?;
+                } else {
+                    write!(w, "Some(crate::Shape::from(&[{}])", dims.join(", "))?;
+                }
+            }
+            Some(AttrValue_oneof_value::tensor(tensor)) => {
+                dbg!(tensor);
```
adamcrume, line 485: There's a few debug statements that need to be removed.

adamcrume (changes requested):
I'd like to share common code between tensorflow-op-codegen/src/main.rs and tensorflow-op-codegen/src/bin/eager.rs, but since I forgot about this PR for a while (sorry about that!), I can clean that up later.

tensorflow-op-codegen/src/bin/eager.rs
dskkato: `dbg!`s were removed.

dskkato:
There is a lot of duplicate code as you have pointed out. I was originally trying to see if I could handle this with only minor changes, but it was difficult to deal with the small (but many) differences in the generated code, so I copied most of them.

adamcrume (approved):

adamcrume:
Thanks!

dskkato:
Please note the code generation was performed for older TensorFlow, so please re-run it if needed.

PR #330, opened 2021-11-11
Fix the build status badge link from travis to GitHub Actions

dskkato:
This is a correction of what was left undone in #327.


adamcrume:
Thanks!

PR #332, opened 2021-11-13
Add the eager api ops wrapper.

dskkato:
In this PR, I added some basic wrappers required to construct raw_ops later on, and omitted utility-like functions which might be useful to investigate Op's internal state.

According to the documentation, TFE_Op is very similar toTF_OperationDescription with some differences:
```
// (1) TF_Output or TFE_TensorHandle* as arguments to TF_AddInput,
//     TF_AddInputList
// (2) TF_ColocateWith, TF_AddControlInput etc. do not make sense.
// (3) Implementation detail: Avoid use of NodeBuilder/NodeDefBuilder since
//     the additional sanity checks there seem unnecessary;
```

@adamcrume I think this PR is now ready for review. Please review this when you have time. Please note that raw_ops is just a rough sketch and actual implementation will be added via `tensorflow-op-codegen` like `ops_impl.rs`.

This patch is a smaller split of #326, and is related to #321.


src/eager/raw_ops.rs
```diff
@@ -0,0 +1,82 @@
+/// Tentative implementation of raw_ops for unit testing.
```
adamcrume, line 1: If this is only for unit testing, then the module should be hidden behind a `#[cfg(test)]`, and the name should somehow reflect that it contains test helpers.

src/eager/tensor_handle.rs
```diff
@@ -221,6 +221,15 @@ impl<'a> TensorHandle<'a> {
             }
         }
     }
+
+    /// Convert the raw TFE_TensorHandle* into a TensorHandle.
+    #[allow(dead_code)] // until raw_ops are implemented
```
adamcrume, line 226: It looks like this isn't necessary.

src/eager/tensor_handle.rs
```diff
@@ -221,6 +221,15 @@ impl<'a> TensorHandle<'a> {
...
+    pub(super) fn from_tensor_handle(_ctx: &'a Context, inner: *mut tf::TFE_TensorHandle) -> Self {
```
adamcrume, line 227: Since this takes raw pointers, it should be an `unsafe` function.

adamcrume (changes requested):

src/eager/tensor_handle.rs
dskkato: marked as `unsafe` now.

src/eager/tensor_handle.rs
dskkato: Removed. Thanks,

src/eager/op.rs
```diff
@@ -0,0 +1,532 @@
...
+    }
+
+    pub(super) fn get_name(&self) -> Result<&str> {
+        let status = Status::new();
+
+        let name = unsafe {
+            let name = tf::TFE_OpGetName(self.inner, status.inner);
+            CStr::from_ptr(name)
+        };
+        if status.is_ok() {
```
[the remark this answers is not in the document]
dskkato: Updated. Thank you for your advice.

src/eager/op.rs
```diff
@@ -0,0 +1,532 @@
...
+            ctx: PhantomData,
+        })
+    }
+
+    pub(super) fn get_name(&self) -> Result<&str> {
+        let status = Status::new();
+
+        let name = unsafe {
+            let name = tf::TFE_OpGetName(self.inner, status.inner);
+            CStr::from_ptr(name)
```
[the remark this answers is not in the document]
dskkato: Fixed.

src/eager/raw_ops.rs
dskkato: I've moved them down one level to make it more obvious that they are helper functions for Op unit tests. I will also put raw_ops, which will be added later, in this position.

dskkato:
I set the scope of the Op to the narrowest, keeping in mind that I will only expose the raw_ops operations in the future.

Exposing the Op structure would allow users of this crate to implement the raw_ops equivalent freely, but since the API is still immature, I thought it would be better to hide it as an implementation detail.

adamcrume:
Looks good.  It'll be ready to merge once `get_output_length` and `get_input_length` are removed.  They can always be added back later if needed.

src/eager/op.rs
```diff
@@ -0,0 +1,532 @@
...
+                value.inner()?,
+                status.inner(),
+            );
+        }
+        status.into_result()
+    }
+
+    /// May be useful for debug, but not sure how to use it now.
+    fn get_output_length(&self) -> usize {
+        unimplemented!();
```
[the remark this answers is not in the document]
dskkato: These unimplemented wrappers were removed.

dskkato:
Those wrappers are removed now.

adamcrume:
Thanks!  Lots of good progress. :)

PR #334, opened 2021-12-07
fix(tensorflow-sys build): Fix deprecated

gaocegege:
```rust
    /// Get the name of the file in a sanitized form. It truncates the name to the first NULL byte,
    /// removes a leading '/' and removes '..' parts.
    #[deprecated(
        since = "0.5.7",
        note = "by stripping `..`s from the path, the meaning of paths can change.
                `mangled_name` can be used if this behaviour is desirable"
    )]
    pub fn sanitized_name(&self) -> ::std::path::PathBuf {
        self.mangled_name()
    }
```

The func sanitized_name is deprecated since 0.5.7.

### DIFF
--- a/tensorflow-sys/build.rs
+++ b/tensorflow-sys/build.rs
@@ -150,7 +150,7 @@ fn extract_zip<P: AsRef<Path>, P2: AsRef<Path>>(archive_path: P, extract_to: P2)
     let mut archive = ZipArchive::new(file).unwrap();
     for i in 0..archive.len() {
         let mut zipfile = archive.by_index(i).unwrap();
-        let output_path = extract_to.as_ref().join(zipfile.sanitized_name());
+        let output_path = extract_to.as_ref().join(zipfile.mangled_name());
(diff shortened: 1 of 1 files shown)


tensorflow-sys/build.rs
```diff
@@ -150,7 +150,7 @@ fn extract_zip<P: AsRef<Path>, P2: AsRef<Path>>(archive_path: P, extract_to: P2)
     let mut archive = ZipArchive::new(file).unwrap();
     for i in 0..archive.len() {
         let mut zipfile = archive.by_index(i).unwrap();
-        let output_path = extract_to.as_ref().join(zipfile.sanitized_name());
+        let output_path = extract_to.as_ref().join(zipfile.mangled_name());
```
adamcrume, line 153: It looks like `enclosed_name` is preferred over `mangled_name`.

gaocegege:
Sorry I do not have the bandwidth to fix the PR, I am closing it.

PR #336, opened 2021-12-17
`private-docs-rs` feature should be propagated to tensorflow-sys.

### LINKED ISSUE #255: docs.rs broken
See https://groups.google.com/a/tensorflow.org/forum/#!topic/rust/VBcTPuas3sM.  We may need to use the approach in https://github.com/rust-lang/docs.rs/issues/806#issuecomment-638355488.

### LINKED ISSUE #335: Test that docs build on CI
Currently the docs are broken https://docs.rs/crate/tensorflow/0.17.0

Idk what the issue is but it should be noted that docs.rs uses nightly. You can look at a hopefully properly implemented GH action for example [here](https://github.com/martin-t/rg3d/blob/ci-fixes/.github/workflows/ci.yml#L101)

dskkato:
The private-docs-rs feature was introduced to avoid large file downloading when building docs in docs.rs. However, this trigger was not propagated to `tensorflow-sys` which have broken [the tensorflow doc in docs.rs](https://docs.rs/crate/tensorflow/0.17.0). This patch is trying to resolve the following issues:

Resolves #255
Resolves #335

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -43,7 +43,7 @@ tensorflow_unstable = []
 eager = ["tensorflow-sys/eager"]
 # This is for testing purposes; users should not use this.
 examples_system_alloc = ["tensorflow-sys/examples_system_alloc"]
-private-docs-rs = [] # DO NOT RELY ON THIS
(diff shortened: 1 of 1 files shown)


adamcrume:
Hm, good catch.  I hope this fixes it.

PR #338, opened 2021-12-25
Introduce ToTensorHandle trait and its implementations. 

dskkato:
This pull request introduces the ToTensorHandle trait to make the API for eager execution easier to handle.

I also tried to implement this trait for some types other than Tensor/TensorHandle. However, since this is before the implementation of raw_opw, it may be better to implement other types than Tensor/TensorHandle later.

### DIFF
--- a/src/eager.rs
+++ b/src/eager.rs
@@ -15,6 +15,9 @@ pub use op::raw_ops;
 
 use crate::{Result, Tensor, TensorType};
 
+#[cfg(feature = "ndarray")]
+use ndarray::{ArrayBase, Data, Dimension};
+
 /// Simple helper trait to convert a Tensor into a TensorHandle for use in eager
 /// execution.
 pub trait ToTensorHandle<'a> {
@@ -60,11 +63,31 @@ impl<'a, T: TensorType> ToTensorHandle<'a> for [T] {
     }
 }
 
+#[cfg(feature = "ndarray")]
+/// Convert any ndarray::ArrayBase type into a TensorHandle
+impl<'a, T, S, D> ToTensorHandle<'a> for ArrayBase<S, D>
+where
+    T: TensorType,
+    S: Data<Elem = T>,
+    D: Dimension,
+{
+    fn to_handle(&self, ctx: &'a Context) -> Result<TensorHandle<'a>> {
+        let dims: Vec<u64> = self.shape().iter().map(|x| *x as u64).collect();
+        let mut tensor: Tensor<T> = Tensor::new(&dims);
+        for (e, v) in tensor.iter_mut().zip(self.iter()) {
+            e.clone_from(v);
+        }
+        TensorHandle::new(ctx, &tensor)
+    }
+}
 #[cfg(test)]
 mod tests {
     use super::*;
     use crate::eager::{Context, ContextOptions};
     use crate::Tensor;
+
+    #[cfg(feature = "ndarray")]
+    use ndarray::array;
 
     #[test]
     fn tensor_to_handle() {
@@ -142,5 +165,19 @@ mod tests {
         let handle = v.to_handle(&ctx).unwrap();
         let tensor: Tensor<f64> = handle.resolve().unwrap();
         assert_eq!(&v, &tensor[..]);
+    }
+
+    #[cfg(feature = "ndarray")]
+    #[test]
+    fn ndarray_to_handle() {
+        let a = array![[1i32, 2], [3, 4]];
+
+        let ctx = Context::new(ContextOptions::new()).unwrap();
+        let handle = a.to_handle(&ctx).unwrap();
+        let tensor: Tensor<i32> = handle.resolve().unwrap();
+        assert_eq!(a.as_slice().unwrap(), &tensor[..]);
+
+        let at = Tensor::from(a);
+        assert_eq!(at, tensor);
     }
 }
--- a/src/eager.rs
+++ b/src/eager.rs
@@ -63,6 +63,13 @@ impl<'a, T: TensorType> ToTensorHandle<'a> for [T] {
     }
 }
 
+impl<'a, T: TensorType, const N: usize> ToTensorHandle<'a> for [T; N] {
+    fn to_handle(&self, ctx: &'a Context) -> Result<TensorHandle<'a>> {
+        let tensor = Tensor::from(&self[..]);
+        TensorHandle::new(ctx, &tensor)
+    }
+}
+
 #[cfg(feature = "ndarray")]
 /// Convert any ndarray::ArrayBase type into a TensorHandle
 impl<'a, T, S, D> ToTensorHandle<'a> for ArrayBase<S, D>
--- a/src/eager.rs
+++ b/src/eager.rs
@@ -15,9 +15,6 @@ pub use op::raw_ops;
 
 use crate::{Result, Tensor, TensorType};
 
-#[cfg(feature = "ndarray")]
-use ndarray::{ArrayBase, Data, Dimension};
-
 /// A helper trait to convert a Tensor or some other types into a TensorHandle
 /// for use in eager execution.
 pub trait ToTensorHandle<'a> {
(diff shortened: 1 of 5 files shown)


src/eager.rs
```diff
@@ -12,3 +12,179 @@ pub use tensor_handle::*;
...
+    fn to_handle(&self, ctx: &'a Context) -> Result<TensorHandle<'a>> {
+        let dims: Vec<u64> = self.shape().iter().map(|x| *x as u64).collect();
+        let mut tensor: Tensor<T> = Tensor::new(&dims);
+        for (e, v) in tensor.iter_mut().zip(self.iter()) {
+            e.clone_from(v);
+        }
+        TensorHandle::new(ctx, &tensor)
+    }
+}
+#[cfg(test)]
```
dskkato, line 90: Need a blank line before this line.

src/eager.rs
```diff
@@ -12,3 +12,181 @@ pub use tensor_handle::*;
...
+/// A helper trait to convert a Tensor or some other types into a TensorHandle
+/// for use in eager execution.
+pub trait ToTensorHandle<'a> {
+    /// Convert a Tensor or values into a new TensorHandle.
+    ///
+    /// _Warning_ : This function may create multiple handles to the same
+    /// underlying tensor. Users should be careful not to modify the tensor
+    /// after converting it to a TensorHandle. Also, users should be careful
+    /// not to modify the Tensor generated from the TensorHandle's `resolve`
+    /// method.
```
adamcrume, line 30: I should have caught this sooner, but this doesn't just have the potential to cause user confusion and bugs; it could actually invoke undefined behavior: https://doc.rust-lang.org/reference/behavior-considered-undefined.html.  (It falls under "Breaking the pointer aliasing rules".)  At least, if the data is modified while it's held behind a reference in Rust elsewhere, it's definitely undefined behavior.  With the FFI involved, it's less clear.  I have no idea if e.g. modifying data through a `&mut` and reading it in C++ while the reference is still live causes undefined behavior or not, although it probably is.

The eager mode support needs to guarantee through some mechanism that no safe code can break the pointer aliasing rules.  Unfortunately, the C API doesn't make that easy, since it only says `TFE_TensorHandleResolve` *might* alias internal memory, and there's no clear way to find out what it is or isn't aliasing.  We could do a defensive copy, although that would be less efficient, of course.  Another idea would be to create something like a `ReadonlyTensor` type which looks like `Tensor` but only supports readonly operations.

We also need to take care of `TensorHandle::new`.  It probably has the same fixes (either copy, i.e. take a `Tensor` instead of a `&Tensor`, or pass a `&ReadonlyTensor`).

It also occurs to me that `TensorHandle::new` has lifetime issues:

```rust
let ctx: Context = ...;
let handle = {
  let tensor: Tensor<f32> = ...;
  TensorHandle::new(&ctx, &tensor)?
};
// At this point, the tensor has deallocated its memory, but we still have a handle.
```

That could be fixed by adding a constraint so that the handle outlives both the context and the tensor, e.g.

```rust
impl<'a> TensorHandle<'a> {
    pub fn new<T: TensorType, 'b: 'a, 'c: 'a>(_ctx: &'b Context, t: &'c Tensor<T>) -> Result<TensorHandle<'a>> {
```

An alternative to all of the above is to mark all of these methods as unsafe and require the client to manage this somehow.  That's not a very user-friendly interface, though.

adamcrume (changes requested):

src/eager.rs
dskkato: > I should have caught this sooner, but this doesn't just have the potential to cause user confusion and bugs; it could actually invoke undefined behavior: https://doc.rust-lang.org/reference/behavior-considered-undefined.html. (It falls under "Breaking the pointer aliasing rules".) At least, if the data is modified while it's held behind a reference in Rust elsewhere, it's definitely undefined behavior. With the FFI involved, it's less clear. I have no idea if e.g. modifying data through a `&mut` and reading it in C++ while the reference is still live causes undefined behavior or not, although it probably is.

I have a similar concern, for example, the following code seems to cause non-mutable memory to be modified outside of that variable. I added a unittest to ensure the eager mode API does not break the Rust's memory design. I think this is partially related to what you pointed. Currently, this unittest will fail:

https://github.com/tensorflow/rust/blob/05c49af1b4f3548b765c026d0ddca11662db9e4f/src/eager.rs#L101-L122

> The eager mode support needs to guarantee through some mechanism that no safe code can break the pointer aliasing rules. Unfortunately, the C API doesn't make that easy, since it only says `TFE_TensorHandleResolve` _might_ alias internal memory, and there's no clear way to find out what it is or isn't aliasing. We could do a defensive copy, although that would be less efficient, of course. Another idea would be to create something like a `ReadonlyTensor` type which looks like `Tensor` but only supports readonly operations.

It seems a good idea to create a ReadonlyTensor, but it was not immediately clear to what extent it should be the ReadonlyTensor for the eager execution. I will think about it some more in conjunction with the above problem.

> We also need to take care of `TensorHandle::new`. It probably has the same fixes (either copy, i.e. take a `Tensor` instead of a `&Tensor`, or pass a `&ReadonlyTensor`).
> 
> It also occurs to me that `TensorHandle::new` has lifetime issues:
> 
> ```rust
> let ctx: Context = ...;
> let handle = {
>   let tensor: Tensor<f32> = ...;
>   TensorHandle::new(&ctx, &tensor)?
> };
> // At this point, the tensor has deallocated its memory, but we still have a handle.
> ```

If we can rely on the implementation detail of the TensorFlow's Tensor class, this might be fine because each Tensor is reference counted, and creating new TensorHandle increments the count. So destructing Tensor while leaving only its handles may be fine, although we cannot read out this point from the `eager/c_api.h`. I will consider these points a little more as well, including the following points.

> 
> That could be fixed by adding a constraint so that the handle outlives both the context and the tensor, e.g.
> 
> ```rust
> impl<'a> TensorHandle<'a> {
>     pub fn new<T: TensorType, 'b: 'a, 'c: 'a>(_ctx: &'b Context, t: &'c Tensor<T>) -> Result<TensorHandle<'a>> {
> ```
> 
> An alternative to all of the above is to mark all of these methods as unsafe and require the client to manage this somehow. That's not a very user-friendly interface, though.

dskkato: I introduced ReadonlyTensor at a4111df. The ReadonlyTensor is a Tensor that does not have mutation operations of the underlying memory. The conversion between Tensor and ReadonlyTensor is performed by moving the ownership of the underlying buffers.

__Conversion steps__

* `Tensor` -(`freeze`)-> `ReadonlyTensor` -(`to_handle`)-> `TensorHandle`
* `TensorHandle` -(`resolve`)-> `ReadonlyTensor` -(`into_tensor`)-> `Tensor`

The unsafe point is a conversion from ReadonlyTensor to Tensor, which will result in "Breaking the pointer aliasing rules" when modified. So it is marked as unsafe (the `into_tensor` method).

Ideally, it would be more useful if the ReadonlyTensor could be fed into the `SessionRunArgs` directly, but it is not done since it needs more work (maybe separating Tensor's write/read interface).

dskkato:
https://github.com/tensorflow/tensorflow/blob/bd9ba9309d28bc48675c029b60f45156b0c78686/tensorflow/c/eager/c_api_test_util.cc#L97-L109

src/eager.rs
dskkato: > > We also need to take care of `TensorHandle::new`. It probably has the same fixes (either copy, i.e. take a `Tensor` instead of a `&Tensor`, or pass a `&ReadonlyTensor`).
> > It also occurs to me that `TensorHandle::new` has lifetime issues:
> > ```rust
> > let ctx: Context = ...;
> > let handle = {
> >   let tensor: Tensor<f32> = ...;
> >   TensorHandle::new(&ctx, &tensor)?
> > };
> > // At this point, the tensor has deallocated its memory, but we still have a handle.
> > ```
> 
> If we can rely on the implementation detail of the TensorFlow's Tensor class, this might be fine because each Tensor is reference counted, and creating new TensorHandle increments the count. So destructing Tensor while leaving only its handles may be fine, although we cannot read out this point from the `eager/c_api.h`. I will consider these points a little more as well, including the following points.
> 

I checked this point in [the unit test code](https://github.com/tensorflow/tensorflow/blob/c256c071bb26e1e13b4666d1b3e229e110bc914a/tensorflow/c/eager/c_api_test_util.cc#L83-L95). It seems that deleting TF_Tensor while leaving the corresponding TFE_TensorHandle is the valid use case.

```c++
TFE_TensorHandle* DoubleTestMatrixTensorHandle(TFE_Context* ctx) {
  int64_t dims[] = {2, 2};
  double data[] = {1.0, 2.0, 3.0, 4.0};
  TF_Status* status = TF_NewStatus();
  TF_Tensor* t = TFE_AllocateHostTensor(ctx, TF_DOUBLE, &dims[0],
                                        sizeof(dims) / sizeof(int64_t), status);
  memcpy(TF_TensorData(t), &data[0], TF_TensorByteSize(t));
  TFE_TensorHandle* th = TFE_NewTensorHandleFromTensor(ctx, t, status);
  CHECK_EQ(TF_OK, TF_GetCode(status)) << TF_Message(status);
  TF_DeleteTensor(t);
  TF_DeleteStatus(status);
  return th;
}
```
I added this information in the TensorHandle's docs.

src/eager/readonly_tensor.rs
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    /// Returns the tensor's dimensions.
+    pub fn dims(&self) -> &[u64] {
+        &self.dims
+    }
+
+    /// Returns the tensor's dimensions as a Shape.
+    pub fn shape(&self) -> Shape {
+        Shape(Some(self.dims.iter().map(|d| Some(*d as i64)).collect()))
```
adamcrume, line 95: I see you copied this from Tensor, but it can be written more succinctly nowadays as `Shape::from(&self.dims)`

src/eager.rs
adamcrume: Interesting.  I didn't realize they were using reference counting.  We may also need a safe (i.e. not `unsafe`) function to copy a `ReadonlyTensor` to a new `Tensor` and a trait to abstract the common behavior for `Tensor` and `ReadonlyTensor`, but we can always add those later as needed.  This looks good for now.

adamcrume (changes requested):

src/eager/readonly_tensor.rs
dskkato: Thanks. It seems that I need to pass a slice of the dims explicitly (deref coercion did not work here).

adamcrume:
Thanks!

PR #344, opened 2022-01-12
Add ToTensorHandle implementations for other types.

dskkato:
In this PR, I have implemented the ToTensorHandle trait for several data types that can be converted to Tensor to make the eager API easier to use.

This is a part of  #321.


adamcrume:
Thanks!

PR #351, opened 2022-03-30
Add support for static linking

sgrif:
While tensorflow doesn't officially support building a static library,
it is relatively straightforward on Unix using instructions found by the
community at https://github.com/tensorflow/tensorflow/issues/28388. This
allows tensorflow-sys to use a static library should one already exist.

No attempt is made to add any support for automatically building a
static library. It is entirely on the user to go through the trouble of
producing `libtensorflow.a` should they want to use it. This commit
*only* allows the user to ask the build script to use the static library
they have built.

### DIFF
--- a/tensorflow-sys/README.md
+++ b/tensorflow-sys/README.md
@@ -60,6 +60,9 @@ compiled library will be picked up.
 **macOS Note**: Via [Homebrew](https://brew.sh/), you can just run
 `brew install libtensorflow`.
 
+To statically link Tensorflow, set the environment variable
+`TENSORFLOW_LIB_STATIC` to the directory containing `libtensorflow.a`
+
 ## Resources
 
 [bazel]: http://www.bazel.io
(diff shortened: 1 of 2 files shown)


adamcrume:
Can you make sure the checks all pass?

sgrif:
Whoops, looks like I forgot to run `cargo fmt`. I will make sure to run it first thing Monday morning. The PR should otherwise be fine to review in the meantime

sgrif:
The checks should pass now but I'm unable to rerun them

adamcrume (changes requested):
Can you run `./test-all` while building statically?  When I tried, it crashed with: `Non-OK-status: status status: NOT_FOUND: Op type not registered 'NoOp'`.

sgrif:
I left the job where this was desirable and no longer have the machine I was using to work on this. So someone else will have to pick this up to make this happen.

Sorry for wasting your time

adamcrume:
I'm sorry to hear that, and I'm sorry I didn't review this earlier.  I hope your job situation is okay!

PR #353, opened 2022-04-13
Generate eager and ndarray docs when releasing.

dskkato:
Thank you for releasing v.0.18.0 of this crate. Also, the documentation for this crate is back in #336 and users can now check it at docs.rs.

https://docs.rs/tensorflow/0.18.0/tensorflow/index.html

However, I noticed that there is no additional features when generating the documentation. Also, the docs for https://tensorflow.github.io/rust/tensorflow/ are missing the eager and ndarray feature docs.

This PR resolves the above inconsistent docs in both docs.rs and github.io.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,7 @@ doctest = false
 
 # Prevent downloading or building TensorFlow when building docs on docs.rs.
 [package.metadata.docs.rs]
-features = ["private-docs-rs"]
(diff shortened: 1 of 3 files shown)


adamcrume:
Thanks!  This was an oversight.  I wish that rustdoc would automatically indicate when additional features are required for certain items, but at least documenting it ourselves should be sufficient for now.

PR #354, opened 2022-04-13
Adding the ability to link at runtime

brianjjones:
This change will allow you to build this crate
using runtime linking. Users will need to call
tensorflow::library::load() before any other calls.


brianjjones:
I'm looking to add TensorFlow as a backend option for wasi-nn (https://github.com/bytecodealliance/wasi-nn). We'd like to support runtime linking, because in some cases TensorFlow installation will be optional. In these cases the application may want to probe the system to see if TensorFlow is available before running an inference with the library. If its not, the application can choose a different machine learning framework or return an informative error message. With dynamic linking, the application simply won't load if the library isn't available, which is something we don't want to happen.

adamcrume:
Can you make sure the checks all pass?

brianjjones:
Sorry for the delay, I fixed the formatting so it would pass CI.

adamcrume (changes requested):

src/graph.rs
```diff
@@ -640,6 +640,7 @@ impl Graph {
     /// # Returns
     ///
     ///  A newly created `Function` instance.
+    #[cfg(not(feature = "tensorflow_runtime_linking"))]
```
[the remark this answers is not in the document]
brianjjones: It turns out it was a limitation of Rust in regards to the Debug macro, if a function pointer has more than 12 arguments it fails. Rather than removing these functions I'll be removing the #[define(debug)] in runtime.rs since there isn't really a reason for it.

brianjjones:
Thanks for the feedback, I think I addressed all the issues and fixed a few other minor things.

tensorflow-sys/src/runtime_linking/runtime.rs
```diff
@@ -0,0 +1,114 @@
...
+        /// Load all of the function definitions from a shared library.
+        pub fn load() -> Result<(), String> {
+            match crate::library::find() {
+                None => Err("Unable to find the `libtensorflow` library to load".into()),
+                Some(path) => load_from(path),
+            }
+        }
+        fn load_from(path: PathBuf) -> Result<(), String> {
+            let library = Arc::new(SharedLibrary::load(path)?);
+            *LIBRARY.write().unwrap() = Some(library);
```
adamcrume, line 79: Since this is returning a `Result` anyway, this should return an `Err` on failure rather than calling `unwrap()` and panicking.

adamcrume (changes requested):
One small change and this should be good to go.

tensorflow-sys/src/runtime_linking/runtime.rs
brianjjones: Good point, updated to fix this.

brianjjones: I misunderstood what was being asked apparently and have updated the solution.

adamcrume: I meant that we shouldn't be using `unwrap()`, so instead of

```rust
let mut static_lib = LIBRARY.write().map_err(...).unwrap();
```

it should be

```rust
let mut static_lib = LIBRARY.write().map_err(...)?;
```

adamcrume:
Thanks!

PR #355, opened 2022-04-14
update mobilenetv3 example.

dskkato:
Now an input image is loaded using the eager API, so the `image` dev-dependency was removed.


adamcrume:
Thanks!

PR #363, opened 2022-07-01
Improve debug and printing of `Tensor` and `ReadonlyTensor`

jackos:
This adds debug and alternate debug of the values in a tensor, which improves dev experience when experimenting with operations, expressions, and eager raw_ops on tensors.

### Standard Debug
```rust
let values: Vec<u16> = (1..9).collect(); 
let tensor = Tensor::new(&[2, 2, 2]).with_values(&values)?;

println!("{tensor:?}");
```
```plaintext
Tensor<u16> { values: [[[1, 2], [3, 4]], [[5, 6], [7, 8]]], dtype: uint16, shape: [2, 2, 2]}
```
### Alternate Debug (pretty)
```rust
dbg!(tensor);
```
```plaintext
[examples/show.rs:10] &tensor = Tensor<u16> {
    values: [
        [[1, 2], [3, 4]],
        [[5, 6], [7, 8]]
    ],
    dtype: uint16,
    shape: [2, 2, 2]
}
```
It also adds a maximum to how many elements are printed (default 1000 like Python), which works across printing and debugging to only display a `...` if the maximum elements are exceded:
### Print max elements exceeded
```bash
TF_RUST_DISPLAY_MAX=5
```
```rust
dbg!(tensor);
```
```plaintext
[examples/show.rs:11] &tensor = Tensor<u16> {
    values: [
        [[1, 2], [3, 4]],
        [[5, ...], ...]
    ],
    dtype: uint16,
    shape: [2, 2, 2]
}
```

```rust
println!("{tensor}");
```
```plaintext
[[[1, 2], [3, 4]], [[5, ...], ...]]
```


adamcrume:
This looks useful, but I'm not convinced that truncating the tensors is the best behavior.  `println!("{:?}, vec![0.0; 10000])` prints everything, for example.

jackos:
> This looks useful, but I'm not convinced that truncating the tensors is the best behavior. `println!("{:?}, vec![0.0; 10000])` prints everything, for example.

Do you mean just for printing or debugging as well?

The intention was to follow the way the Python version works by guarding against a huge amount of data being printed to someone unsuspecting, but if you want it can be defaulted to off and turned on via setting a limit on the environment variable? Makes sense to follow normal Rust behavior.

src/eager/readonly_tensor.rs
```diff
@@ -32,6 +36,38 @@ impl<T: TensorType> Deref for ReadonlyTensor<T> {
     }
 }
 
+impl<T: TensorType> Debug for ReadonlyTensor<T> {
```
adamcrume, line 39: ReadonlyTensor could use a Display impl, too, to keep it consistent with Tensor.

src/eager/readonly_tensor.rs
```diff
@@ -32,6 +36,38 @@ impl<T: TensorType> Deref for ReadonlyTensor<T> {
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+        let mut counter: i64 = match std::env::var("TF_RUST_DISPLAY_MAX") {
+            Ok(e) => e.parse().unwrap_or(-1),
+            Err(_) => -1,
+        };
+        let dtype = self.data_type().to_string().to_lowercase();
```
adamcrume, line 45: There's no need to convert it to lowercase.  The case will match the cases of the DataType enum.

src/lib.rs
```diff
@@ -2377,6 +2444,108 @@ mod tests {
         }
     }
 
+    #[test]
+    fn tensor_debug() {
```
adamcrume, line 2448: We also need to test scalar tensors and zero-element tensors.

src/lib.rs
```diff
@@ -2377,6 +2444,108 @@ mod tests {
...
+        let values: Vec<i32> = (1..9).collect();
+        let tests = [
+            (
+                "Tensor<i32> { values: [1, 2, 3, 4, 5, 6, 7, 8], dtype: int32, shape: [8]}",
```
adamcrume, line 2452: This format has spaces around the opening curly brace, but not the last one, which is inconsistent.

src/lib.rs
```diff
@@ -2377,6 +2444,108 @@ mod tests {
...
+        let pretty = "[
+        1,
+        2,
+        3,
+        4,
+        5,
+        6,
+        7,
+        8";
+        test_debug_i32_pretty(values, shape, pretty)
```
adamcrume, line 2496: It'd be a lot easier to verify that these tests are testing the right thing if `pretty` contained the entire string verbatim, i.e. inline `test_debug_i32_pretty` and don't format or concatenate the expected strings in the tests.

src/lib.rs
```diff
@@ -2377,6 +2444,108 @@ mod tests {
...
+        let shape = &[5, 2, 2][..];
+
+        let pretty = "[
+        [[1, 2], [3, 4]],
+        [[5, 6], [7, 8]],
+        [[9, 10], ...],
+        ...";
+        test_debug_i32_pretty(values, shape, pretty);
+    }
+
```
adamcrume, line 2548: The "pretty" format also needs to test scalar and zero-element tensors.

adamcrume (changes requested):

jackos:
@adamcrume thanks for the review, have fixed all of your comments

src/lib.rs
```diff
@@ -2377,6 +2478,149 @@ mod tests {
...
+        let tensor = Tensor::from(1);
+        let pretty = "Tensor<i32> {\n    scalar_value: 1,\n    dtype: Int32,\n    shape: []\n}";
+        let standard = "Tensor<i32> { scalar_value: 1, dtype: Int32, shape: [] }";
+        assert_eq!(pretty, format!("{tensor:#?}"));
+        assert_eq!(standard, format!("{tensor:?}"));
+    }
+
+    #[test]
+    fn tensor_debug_zero_elements() {
+        let tensor = Tensor::<f32>::new(&[0, 0]);
```
adamcrume, line 2519: To test for the bug I mentioned above, this could use a shape like [0, 10].

adamcrume (changes requested):

adamcrume:
Just a few minor issues to resolve, and this'll be ready.

src/lib.rs
```diff
@@ -1260,12 +1260,142 @@ impl<T: TensorType> Clone for TensorDataNoCRepr<T> {
...
+) -> fmt::Result {
+    let mut counter: i64 = match std::env::var("TF_RUST_DISPLAY_MAX") {
+        Ok(e) => e.parse().unwrap_or(-1),
+        Err(_) => -1,
+    };
+    let values_name = match dims.len() {
+        0 => "scalar_value",
+        _ => "values",
+    };
+    let elements: u64 = dims.iter().sum();
```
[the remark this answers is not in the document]
jackos: Great point didn't think about that, I went with `if tensor.is_empty()`

jackos:
> Just a few minor issues to resolve, and this'll be ready.
Great thanks fixed the errors

adamcrume:
Sorry for the delay!  Looks good, and thanks!

PR #366, opened 2022-07-14
Enhancement 284 Sealed Trait

### LINKED ISSUE #284: Should TensorType be a sealed trait or marked unsafe?
Hi there, we (Rust group @sslab-gatech) are scanning crates on crates.io for potential soundness bugs. We noticed that the [`TensorType`](https://tensorflow.github.io/rust/tensorflow/trait.TensorType.html) trait is commented as:

> Currently, all implementors must not implement Drop (or transitively contain anything that does) and must be bit-for-bit compatible with the corresponding C type. Clients must not implement this trait.

Since implementing this trait improperly potentially allows for dangerous behavior and is commented as something that shouldn't be implemented, would it make sense to seal this trait as per https://rust-lang.github.io/api-guidelines/future-proofing.html#sealed-traits-protect-against-downstream-implementations-c-sealed or to mark it an `unsafe trait`?

eaeiv:
Adds generic `Sealed` super trait and applies it to `TensorType`.

Fixes #284

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -644,11 +644,26 @@ pub type Result<T> = std::result::Result<T, Status>;
 
 ////////////////////////
 
+/// A common implementation of the sealed supertrait programming pattern 
+/// (C-SEALED)
+/// 
+/// With this, implementations of `Sealed` are guarenteed to only exist in the
+/// current crate. This allows modifictation of the underyling traits in ways
+/// that would ordinarily be a breaking change for traits that are not sealed.
+/// 
+/// See https://rust-lang.github.io/api-guidelines/future-proofing.html#sealed-traits-protect-against-downstream-implementations-c-sealed
+ 
+mod private {
+    pub trait Sealed {}
+    
+    impl <T> Sealed for T{}
+}
+
 /// A Rust type that maps to a `DataType`.
 ///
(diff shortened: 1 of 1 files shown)


adamcrume:
Please run `cargo fmt` so the tests pass.  You can test locally by running `./test-all`.

src/lib.rs
```diff
@@ -644,16 +644,31 @@ pub type Result<T> = std::result::Result<T, Status>;
...
+/// With this, implementations of `Sealed` are guarenteed to only exist in the
+/// current crate. This allows modifictation of the underyling traits in ways
+/// that would ordinarily be a breaking change for traits that are not sealed.
+/// 
+/// See https://rust-lang.github.io/api-guidelines/future-proofing.html#sealed-traits-protect-against-downstream-implementations-c-sealed
+ 
+mod private {
+    pub trait Sealed {}
+    
+    impl <T> Sealed for T{}
```
y21, line 659: I don't think this blanket impl makes sense, because it implements Sealed for all types, even for types defined in other crates. That makes the `TensorType: private::Sealed` bound useless, because it's implemented for every type already.

adamcrume: Yep, please fix.

Issue #378, opened 2022-09-11
Migrate from `random` crate to `rand`


adamcrume:
A few examples use the `random` crate and should be using `rand` instead.

PR #379, opened 2022-09-12
migrate from random crate to rand

RealAdarsh:
Migrate from random crate to rand #378

### DIFF
--- a/examples/expressions.rs
+++ b/examples/expressions.rs
@@ -1,4 +1,4 @@
-extern crate random;
+extern crate rand;
 extern crate tensorflow;
 
 use std::error::Error;


adamcrume:
Can you fix the test failures?  The `./test-all` script should succeed.

RealAdarsh:
Hello Adam, I am new to open-source, Could you please help me how to fix this? I could try fixing it.

adamcrume:
All the other places in the file that reference the `random` crate also need to be switched to the `rand` crate, and `Cargo.toml` needs to be updated.  This isn't a simple find-and-replace, because the types and functions in the two crates aren't identical.

adamcrume:
This was fixed in https://github.com/tensorflow/rust/pull/384.

PR #384, opened 2022-10-10
Migrate from `random` to `rand`

fndejan:
Hi, I can see this issue (https://github.com/tensorflow/rust/issues/378) is still open, so here is my PR.


adamcrume:
Sorry for the delay, and thank you!

PR #395, opened 2023-02-11
Upgrade to TensorFlow 2.11.0

adamcrume:


dskkato:
@adamcrume Is it safe to raise the TensorFlow version upto 2.11.0? Windows (native) support (only installed by pip?) should have been discontinued at 2.10.0, but the build configuration seems to have been updated.

https://www.tensorflow.org/install/source_windows#gpu

I don't have a windows machine so I can't try it, but will GPUs continue to work with Windows (native) after upgrading to 2.11.0?

PR #399, opened 2023-02-21
Add support for saving to and restoring from checkpoints

ramon-garcia:
This pull request contributes support for saving and restoring of variables using tensorflow `SaveV2`/`RestoreV2` ops, and therefore it is compatible with checkpoints created with `tf.keras.Model.save_weights`.

It includes a test case.

### DIFF
--- a/src/checkpoint.rs
+++ b/src/checkpoint.rs
@@ -1,0 +1,284 @@
+//! This module supports saving and restoring variables using Tensorflow checkpoints in SaveV2 format
+
+use crate::{Operation, ops, Scope, Session, SessionRunArgs, Status, Tensor};
+use crate::option_insert_result::get_or_insert_with_result;
+
+#[derive(Debug)]
+struct SaveRestoreOps {
+    pub prefix_save: Operation,
+    pub prefix_restore: Operation,
+    pub save_op: Operation,
+    pub restore_op: Operation
+}
+
+/// Checkpointing and restoring struct
+#[derive(Debug)]
+pub struct CheckpointMaker {
+    scope: Scope,
+    variables: Box<[String]>,
+    save_restore_ops: Option<SaveRestoreOps>
+}
+
+impl CheckpointMaker {
+    /// Creates a new CheckpointMaker for a Scope, with a list of variables to save/restore.
+    /// The scope is used to modify the graph to add the save and restore ops.
+    ///
+    /// In order to provide a scope for the CheckpointMaker one can use scope.new_sub_scope("")
+    /// as Scope does not support the Clone trait at present
+    fn new(scope: Scope, variables: Box<[String]>) -> CheckpointMaker {
+        CheckpointMaker { scope, variables, save_restore_ops: None }
+    }
+
+    /// Add save and restore ops to the graph
+    fn build_save_ops(&mut self) -> Result<SaveRestoreOps, Status> {
+        let mut all_variable_ops_opt: Option<Vec<Operation>>  = None;
+        fn make_all_variable_ops(myself: &mut CheckpointMaker) -> Result<Vec<Operation>, Status>
+            {
+                let graph = myself.scope.graph();
+                Ok(myself.variables
+                    .iter()
+                    .map(|v: &String| -> Result<Operation, Status>
+                        { Ok(graph.operation_by_name_required(v.as_str())?.clone()) })
+                    .collect::<Result<Vec<_>, Status>>()?)
+            }
+
+        let existing_save_op = self.scope.graph().operation_by_name("save")?;
+        let (prefix_save, save_op) = if let Some(op) = existing_save_op {
+            let prefix_save_op = self.scope.graph().operation_by_name_required("prefix_save")?;
+            (prefix_save_op, op)
+        } else {
+            let all_variable_ops =
+                get_or_insert_with_result(&mut all_variable_ops_opt, || make_all_variable_ops(self))?;
+            let prefix_save = ops::Placeholder::new()
+                .dtype(crate::DataType::String)
+                .build(&mut self.scope.with_op_name("prefix_save"))?;
+            let tensor_names = ops::constant(
+                &self.variables
+                    .iter()
+                    .map(|v| (*v).to_string())
+                    .collect::<Vec<_>>()[..],
+                &mut self.scope,
+            )?;
+            let shape_and_slices = ops::constant(
+                &self.variables.iter().map(|_| "".to_string()).collect::<Vec<_>>()[..],
+                &mut self.scope,
+            )?;
+            let tensors = all_variable_ops.iter().map(|v|  v.output(0).clone()).collect::<Vec<_>>();
+
+            let mut g = self.scope.graph_mut();
+            let mut nd = g.new_operation("SaveV2", "save")?;
+            nd.add_input(prefix_save.clone());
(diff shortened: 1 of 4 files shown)


ramon-garcia:
A point for discussion: for my own code, I use anyhow crate to have context in exceptions. Otherwise, one gets an error code without any context.

I didn't add anyhow here to avoid add a new dependency to the project, and to be consistent with the rest of this library. But error messages without context are very difficult to diagnose.

adamcrume:
I don't think `anyhow` is appropriate, because it's intended for application code and forces you to use its error type.  That would be awkward for a library, and switching to it would be a breaking change (since it would change the return type of functions).

I'm definitely in favor of including context in errors.  We can manually add context in error messages; it's just a matter of doing it.  To make it easier, we could add

```
impl Status {
  pub fn context(&self, msg: &str) -> Status {
    ...
  }
}
```

and use it with

```
my_tensorflow_function().map_err(|s| s.context("foo"))?;
```

To make this even more concise, we could do the same thing `anyhow` does and provide an extension trait for `Result`, but if we were to expose it outside this crate, we'd probably want to call the methods something other than `context` so they don't conflict for people using both `tensorflow` and `anyhow`.

Changes like this to `Status` should probably be its own PR, though.

ramon-garcia:
Thank you for your reply. Sounds good to me.

src/checkpoint.rs
```diff
@@ -0,0 +1,370 @@
...
+                nd.add_input(tensor_names);
+                nd.add_input(shape_and_slices);
+                let dtypes = all_variable_ops
+                    .iter()
+                    .map(|v| v.get_attr_type("dtype"))
+                    .collect::<Result<Vec<_>, Status>>()?;
+                nd.set_attr_type_list("dtypes", &dtypes[..])?;
+                nd.finish()?
+            };
+            {
```
adamcrume, line 139: This scope doesn't do anything.

src/checkpoint.rs
```diff
@@ -0,0 +1,370 @@
...
+            .shape(var.shape.clone())
+            .build(&mut scope.with_op_name(var.name.as_str()))?;
+        Ok((
+            placeholder.clone(),
+            ops::assign(var.output.clone(), placeholder, scope)?,
+        ))
+    }
+
+    struct MyScopeData {
+        pub scope: Scope,
```
adamcrume, line 239: There's no need to make these `pub`.

src/checkpoint.rs
```diff
@@ -0,0 +1,370 @@
...
+
+#[derive(Debug)]
+struct SaveRestoreOps {
+    pub prefix_save: Operation,
+    pub prefix_restore: Operation,
+    pub save_op: Operation,
+    pub restore_op: Operation,
+}
+
+/// Checkpointing and restoring struct
```
adamcrume, line 14: This documentation needs to be a little more in-depth.

src/checkpoint.rs
```diff
@@ -0,0 +1,370 @@
+//! This module supports saving and restoring variables using Tensorflow checkpoints in SaveV2 format
```
adamcrume, line 1: All the docs need to have punctuation (i.e. end with periods).

adamcrume (changes requested):
Looks great overall, just a few readability concerns.

ramon-garcia:
Thank you for your review. 
In addition, something I am not comfortable with, wouldn't it be better to use an array of variables rather than variable names? When I wrote it, I was concerned about not being able to create a Variable object (for instance, from an imported graph), but it seems to be an independent problem on its own, handled better in its own PR.

src/checkpoint.rs
```diff
@@ -0,0 +1,370 @@
...
+            scope,
+            variables,
+            save_restore_ops: None,
+        }
+    }
+
+    /// Add save and restore ops to the graph
+    fn build_save_ops(&mut self) -> Result<SaveRestoreOps, Status> {
+        let mut all_variable_ops_opt: Option<Vec<Operation>> = None;
+        fn make_all_variable_ops(myself: &mut CheckpointMaker) -> Result<Vec<Operation>, Status> {
```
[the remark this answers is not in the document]
ramon-garcia: The function make_all_variable_ops can be removed if we change from variable names to Variable objects. This is what I am going to implement. Then we can go back if desired.

src/checkpoint.rs
```diff
@@ -0,0 +1,370 @@
+//! This module supports saving and restoring variables using Tensorflow checkpoints in SaveV2 format
+
+use crate::option_insert_result::OptionInsertWithResult;
+use crate::{ops, Operation, Scope, Session, SessionRunArgs, Status, Tensor};
+
+#[derive(Debug)]
+struct SaveRestoreOps {
+    pub prefix_save: Operation,
```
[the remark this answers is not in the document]
ramon-garcia: Done.

src/option_insert_result.rs
```diff
@@ -0,0 +1,18 @@
+// Similar to Option<T>.get_or_insert_with, for a function that returns a result.
+pub trait OptionInsertWithResult<T> {
+    fn get_or_insert_with_result<F, E>(&mut self, f: F) -> Result<&T, E>
```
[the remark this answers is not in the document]
ramon-garcia: Ok, done

src/checkpoint.rs
ramon-garcia: Done

src/checkpoint.rs
ramon-garcia: Ok, I added documentation and a small example at the start of the module.

ramon-garcia:
Added to commits, the first one `013a829`. The next one, `2247ed8` changes the specification of variables from strings to Variable objects. 

Best regards.

ramon-garcia:
Hello?

adamcrume (changes requested):

src/checkpoint.rs
```diff
@@ -1,51 +1,72 @@
-//! This module supports saving and restoring variables using Tensorflow checkpoints in SaveV2 format
-
+//! This module supports saving and restoring variables using Tensorflow checkpoints in SaveV2 format.
+//! First, the user creates a [`CheckpointMaker`] attached to an [`Scope`] indicating the list of variables to be saved/restored.
+//! The CheckpointMaker lazily modifies the graph creating the nodes needed for saving/restoring.
+//! When one wants to save/restore from or into a session, one calls the save/restore  methods
+//! # Example
```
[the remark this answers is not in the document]
ramon-garcia: I hope it is now OK

src/checkpoint.rs
```diff
@@ -1,51 +1,72 @@
...
+/*    fn make_all_variable_ops(&mut self) -> Result<Vec<Operation>, Status> {
+        let graph = self.scope.graph();
+        Ok(self
+            .variables
+            .iter()
+            .map(|v: &String| -> Result<Operation, Status> {
+                Ok(graph.operation_by_name_required(v.as_str())?.clone())
+            })
+            .collect::<Result<Vec<_>, Status>>()?)
+    }*/
```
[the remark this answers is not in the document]
ramon-garcia: Done

ramon-garcia:
ping.

adamcrume:
Sorry for the delay, and thank you!

PR #402, opened 2023-03-04
add PluggableDeviceLibrary

### LINKED ISSUE #381: Add TF_LoadPluggableDeviceLibrary API
Although [`TF_LoadPluggableDeviceLibrary`](https://github.com/tensorflow/tensorflow/blob/master/tensorflow/c/c_api_experimental.cc#L725)  is an experimental API,  it is necessary to use GPU on platform like Mac Arm.

I tried to include this API in code generating but it caused other errors.
Now I always add it manually.

```rust
extern "C" {
    pub fn TF_LoadPluggableDeviceLibrary(
        library_filename: *const ::std::os::raw::c_char,
        status: *mut TF_Status,
    ) -> *mut TF_Library;
}
```

This API had better be included.

dskkato:
Closes #381


dskkato:
Hmmm, how can I test this...?

dskkato:
For mac see #387

dskkato:
directml plugin for windows native may not work. See
https://discuss.tensorflow.org/t/is-it-possible-to-build-tf-with-pluggable-devices-plugin-help-request-pluggable-device/10985/13

src/pluggable_device.rs
```diff
@@ -0,0 +1,53 @@
...
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[ignore]
+    #[test]
+    fn load_pluggable_device_library() {
```
adamcrume, line 47: We might as well leave the test out, since there's no reasonable way to run it (as far as I know).  We have similar issues with TF_LoadLibrary.

adamcrume (changes requested):

src/pluggable_device.rs
dskkato: Yes, right. I tried some, but none of them succeeded. There is too little information on this API.
 
Is there anything else I can do with this pull request?

adamcrume: In https://github.com/tensorflow/rust/pull/387 they solved it (at least for macOS) by installing tensorflow-metal and loading that.  If you want to leave this test out, I can merge this first and ask them to rebase their changes on top of this, so we'd at least have a test that runs in the CI, even if not on everyone's local machine.

dskkato: Thanks for adjusting the PRs. If that order is acceptable, please do so.

A while ago, I was trying to see if the C-API would work with a plugin for Windows, and latest plugin version (currently build from source only) seemed to work with TF 2.12. I'm thinking of trying to see if I can support that as well (apart from this PR) when the version of tensorflow here is updated.

https://github.com/microsoft/tensorflow-directml-plugin

PR #403, opened 2023-03-25
FIXES THE #290

AnakinSkywalkeer:
I added the new enum called AttrValue  and after that i fill the empty field default_value and allowed_values in OpArgDef.

FIXES the #290


AnakinSkywalkeer:
hi @adamcrume i made this pull request can u check it if i need to change smth

adamcrume:
Please ensure that `./test-all` passes.  It looks like the change needs to be formatted.

AnakinSkywalkeer:
@adamcrume hey sorry was busy, so i can u tell what i can use in place  of super::tensor::TensorProtPaeo in Tensor() 
```
#[derive(Debug, Clone, PartialEq)]
pub enum AttrValue {
    S(::std::vec::Vec<u8>),
    I(i64),
    F(f32),
    B(bool),
    FieldType(DataType),
    Shape(Shape),
    Tensor(super::tensor::TensorProtPaeo),
    List(AttrValue_ListValue),
    Func(NameAttrList),
    Placeholder(::std::string::String),
}
```

AnakinSkywalkeer:
it passed the ./test-all

PR #415, opened 2023-09-12
Enable to use newer protobuf

dskkato:


adamcrume:
The problem is that the generated code (which is checked in) is only compatible with the exact same version of `protobuf` that generated it.  This is a very annoying property of the `protobuf` crate, but I can't change that.

We could switch to generating the code in a build script, which means that we need the proto files available at build time.  We could do that by adding TensorFlow as a git submodule, but it's huge (hundreds of megabytes), and apparently cargo doesn't work well with submodules (https://github.com/rust-lang/cargo/issues/7987).  Possibly a better way is to just manually (with a script) copy the protos from TensorFlow into this crate for each version upgrade.

dskkato:
hmmm
 https://github.com/tensorflow/rust/blob/master/src/protos/allocation_description.rs#L1-L1

